PR #11, opened 2020-11-17
Gen2 Color/MonoCamera and StereoDepth example

alex-luxonis:
StereoDepth configuration options:
```
lrcheck  = True   # Better handling for occlusions
extended = False  # Closer-in minimum depth, disparity range is doubled 
subpixel = True   # Better accuracy for longer distance, fractional disparity 32-levels
```

If one or more of the additional depth modes (lrcheck, extended, subpixel) are enabled, then:
 - depth output is FP16. TODO enable U16.
 - median filtering is disabled on device. TODO enable.
 - with subpixel, either depth or disparity has valid data.

Otherwise, depth output is U16 (mm) and median is functional. But like on Gen1, either depth or disparity has valid data. TODO enable both.

Select one pipeline to run:
```
   #pipeline, streams = create_rgb_cam_pipeline()
   #pipeline, streams = create_mono_cam_pipeline()
    pipeline, streams = create_stereo_depth_pipeline()
```

### DIFF
--- a/gen2-camera-demo/main.py
+++ b/gen2-camera-demo/main.py
@@ -177,13 +177,14 @@ def convert_to_cv2_frame(name, image):
         frame = (disp * 255. / max_disp).astype(np.uint8)
 
         # Optionally, apply a color map
-        frame = cv2.applyColorMap(frame, cv2.COLORMAP_HOT)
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon (approved):
LGTM
Tested on BW1092 and also BW1098OBC.  BW1092 images below:
image
image
image
image

PR #12, opened 2020-11-19
Add option to run StereoDepth with input frames from host

alex-luxonis:
Set `source_camera = False`

Using input images from: https://vision.middlebury.edu/stereo/data/scenes2014/

image

image

For the bad looking areas, these are caused by the objects being too close to the camera for the given baseline, exceeding the 96 pixels max distance for disparity matching (StereoDepth engine constraint):
image

These areas will be improved with `extended = True`, however Extended Disparity and Subpixel cannot operate both at the same time.


Luxonis-Brandon (approved):
LGTM
image

PR #24, opened 2020-12-23
Add first-batch of examples

VanDavv:
The first batch of examples created accordingly to https://github.com/luxonis/depthai-docs-website/issues/102

- [x] `01_rgb_preview.py` working fine :heavy_check_mark: 
- [x] `02_mono_preview.py` working fine :heavy_check_mark:
- [x] `03_depth_preview.py` working fine :heavy_check_mark:
- [x] `04_rgb_encoding.py` working fine :heavy_check_mark:
- [x] `05_rgb_mono_encoding.py` working fine :heavy_check_mark:
- [x] `06_rgb_full_resolution_saver.py` working fine :heavy_check_mark:
- [x] `07_mono_full_resolution_saver.py` working fine :heavy_check_mark:
- [x] `08_rgb_mobilenet.py` working fine :heavy_check_mark:
- [x] `09_mono_mobilenet.py` working fine :heavy_check_mark:
- [x] `10_mono_depth_mobilenetssd.py` working fine :heavy_check_mark:


szabi-luxonis:
@VanDavv `04_rgb_encoding.py` and `06_rgb_full_resolution_saver.py` has been fixed.
`06_rgb_full_resolution_saver.py` has very low framerate though, previewout was not meant for such high resolutions.

szabi-luxonis:
`06_rgb_full_resolution_saver.py` is meant to be the same as [this](https://github.com/luxonis/depthai-core/blob/gen2_resource_allocator/examples/src/camera_video_example.cpp) example? Equivalent to `color` stream on gen1

VanDavv:
@szabi-luxonis thanks! 
Just tested and they work indeed, awesome! 
WRT `06_rgb_full_resolution_saver.py` my idea was to show how to see / save the frames in the highest resolution available - as users will be able to decrease it or remove the preview (depending on their needs), the goal is to show the upper limit of what we can achieve

gen2_examples/06_rgb_full_resolution_saver.py
```diff
@@ -0,0 +1,39 @@
...
+from pathlib import Path
+
+import cv2
+import depthai
+import numpy as np
+
+pipeline = depthai.Pipeline()
+
+cam_rgb = pipeline.createColorCamera()
+cam_rgb.setPreviewSize(2160, 3840)
```
szabi-luxonis, line 11: width and height are swapped here

szabi-luxonis (changes requested):

themarpe:
Thanks for all of the examples that are getting into this PR.

2 things I'd like to discuss briefly.

- Should we de facto default to `import depthai as dai` which makes the code look similar in both Python and C++ land?
- I suggest using Device(pipeline) abbreviation instead of the longer:
```
found, device_info = depthai.XLinkConnection.getFirstDevice(depthai.XLinkDeviceState.X_LINK_ANY_STATE)
if not found:
    raise RuntimeError("Device not found")
device = depthai.Device(pipeline, device_info)
```
As it will work correctly in case of 1 device in any state as well as more devices (except in case when all of them are already running - that case I think should throw as well)
The logic behind the default constructor of `Device(pipeline)` is the following: First search for any `UNBOOTED` devices, if not found search for any `BOOTLOADER` devices, and at last search for any `BOOTED` devices.
This also covers the work with 1 device connected to a debugger.

themarpe (approved):
Looks good - feel free to merge
ffmpeg calls are going to be revisited for `depthai-python` examples later on

PR #29, opened 2021-01-07
Add gen2-color-isp-raw experiment

alex-luxonis:
Example app for IMX378 ColorCamera streaming at full resolution / 12MP (4056x3040):
- `isp` - YUV420p uncompressed frames, processed by ISP
- `raw` - BayerBG (R_Gr_Gb_B), as read from sensor, 10-bit packed

One or both streams can be enabled. By default both are enabled (note this slows down FPS, using more USB traffic).

Install the depthai library: `python3 -m pip install -r requirements.txt`
then run `python3 main.py`

User interaction:
- press `c` to capture a set of frames from the enabled streams. Will be saved as `.png` images, as well as YUV420p and unpacked 16bit RAW10 (those can be opened with a viewer like `vooya`).
- press `q` to quit the app.

### DIFF
--- a/gen2-color-isp-raw/main.py
+++ b/gen2-color-isp-raw/main.py
@@ -11,6 +11,7 @@ streams.append('isp')
 
 display_scale = 0.3  # configurable
 
+print("depthai version:", dai.__version__)
 pipeline = dai.Pipeline()
 
 cam = pipeline.createColorCamera()
--- a/gen2-color-isp-raw/main.py
+++ b/gen2-color-isp-raw/main.py
@@ -2,14 +2,36 @@
 
 import cv2
 import numpy as np
+import numba as nb
 import depthai as dai
 
 streams = []
 # Enable one or both streams
 streams.append('isp')
-#streams.append('raw') # TODO: implement unpacking
+streams.append('raw')
 
 display_scale = 0.3  # configurable
+
+''' Packing scheme for RAW10 - MIPI CSI-2
+- 4 pixels: p0[9:0], p1[9:0], p2[9:0], p3[9:0]
+- stored on 5 bytes (byte0..4) as:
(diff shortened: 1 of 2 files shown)


alex-luxonis:
Also works with IMX477 on boards with FFC connector.

gen2-color-isp-raw/main.py
```diff
@@ -0,0 +1,74 @@
...
+streams.append('isp')
+#streams.append('raw') # TODO: implement unpacking
+
+display_scale = 0.3  # configurable
+
+print("depthai version:", dai.__version__)
+pipeline = dai.Pipeline()
+
+cam = pipeline.createColorCamera()
+cam.setEnablePreviewStillVideoStreams(False)
```
themarpe, line 18: Can `raw`, `isp` and other outputs be used concurrently?
Or is this a limitation?
If they can be used, then we can rather do the check for disabling intermediate processing on device side by checking if the output is connected somewhere

alex-luxonis:
Implemented RAW10 unpacking and updated PR description.

gen2-color-isp-raw/main.py
alex-luxonis: Yes, the outputs can be used together with the other post-processed outputs. Disabled them here to reduce the loading on device (and also to avoid the possibility of non-working with some non-standard higher resolutions - around 5K / IMX283).
Definitely we should check inside the node if the outputs are connected, and enable the post-processor accordingly. I just didn't find the API.

themarpe: Okay great - then lets put this as an issue to depthai-device (or core) repository, and we'll pick it up later

szabi-luxonis (approved):
LGTM.

themarpe (approved):

PR #31, opened 2021-01-08
Fix window position

cabelo:


Luxonis-Brandon:
I'm a not great at Github, generally, but were you wanting to merge this @cabelo ?  I noticed that you deleted the branch.

PR #33, opened 2021-01-08
Display two images in one window

cabelo:
Display two images in one window

image


Luxonis-Brandon:
Thanks for this.  Much better idea.  Will give it a test today and then very likely will merge shortly after that.

Luxonis-Brandon (approved):
Thanks for improving this.  Team thinks it's great and fulfills the intention we had with `cv2.moveWindow(...)`.

PR #35, opened 2021-01-10
Update depthai FW; update API calls

szabi-luxonis:


gen2_examples/04_rgb_encoding.py
```diff
@@ -26,7 +26,7 @@
 device.startPipeline()
 
 # Output queue will be used to get the encoded data from the output defined above
-q = device.getOutputQueue(name="h265")
+q = device.getOutputQueue(name="h265", maxSize=4, blocking=False)
```
alex-luxonis, line 29: Better to keep `blocking=True` for the video encoder queues, to not have lost encoded packets -> glitches in the video? And bigger queue size, as those packets are usually small?

 Also in the other examples.

gen2_examples/13_encoding_max_limit.py
```diff
@@ -43,9 +43,9 @@
 dev = dai.Device(pipeline)
 
 # Prepare data queues
-outQ1 = dev.getOutputQueue('ve1Out')
-outQ2 = dev.getOutputQueue('ve2Out')
-outQ3 = dev.getOutputQueue('ve3Out')
+outQ1 = dev.getOutputQueue('ve1Out', maxSize=8, blocking=False)
```
VanDavv, line 46: As @alex-luxonis said, I'd leave blocking=True in the videoEncoder queues

VanDavv (changes requested):
LGTM, but agree with @alex-luxonis on the blobking=True in encoding queues

PR #38, opened 2021-01-12
Update requirements.txt

raudie:
Incorrect depthai version. Not sure if this is the intended version, but the example ran successfully.


Luxonis-Brandon:
Thanks for the catch/fix here.  Double-checking with team.

saching13:
@raudie Thanks for the PR. I think it got updated while rebasing and I didn't notice. Sorry for the trouble.

saching13 (approved):

PR #39, opened 2021-01-12
Add redirect messages to migrated / renamed examples

VanDavv:
As the experiments PR got through - https://github.com/luxonis/depthai-python/pull/77 - and another one, with tweaks, will be merged soon - https://github.com/luxonis/depthai-python/pull/79 - and examples are listed in the gen2 API docs - https://docs.luxonis.com/projects/api/en/docs_python_api/ - we can remove gen2 examples from this repo, as it will be easier to maintain just a single version in depthai-python


themarpe:
Lets remove this once docs cover the description of the examples

Luxonis-Brandon:
Yes, agreed.  And I think we need to keep it as a URL (since we link to it from [KickStarter](https://www.kickstarter.com/projects/opencv/opencv-ai-kit/posts/3068110)), and then just provide the URL from this to the docs page.  

We should always do that also, anything that is linked to, we should try to keep the link alive, with something as simple as a "See our docs _here_ sort of thing".

Thoughts?

szabi-luxonis:
> Yes, agreed. And I think we need to keep it as a URL (since we link to it from [KickStarter](https://www.kickstarter.com/projects/opencv/opencv-ai-kit/posts/3068110)), and then just provide the URL from this to the docs page.
> 
> We should always do that also, anything that is linked to, we should try to keep the link alive, with something as simple as a "See our docs _here_ sort of thing".
> 
> Thoughts?

If we link against commit instead branch: 
e.g.
https://github.com/luxonis/depthai-experiments/blob/0f7fbbdd858b012c999a1cba6f3bd13723dca1bc/gen2-camera-demo/main.py
instead:
https://github.com/luxonis/depthai-experiments/blob/master/gen2-camera-demo/main.py

then the link will be valid and alive until we rewrite the git history.

Luxonis-Brandon:
Oh, good idea.  I will do that from now on.  In this case, KickStarter does not allow editing (after 30 minutes, and the post was days ago) so we will have to keep this around with a link to the docs.

VanDavv:
@Luxonis-Brandon the examples are now in the docs, can we remove them so that we don't have to maintain them in both repositories?

Luxonis-Brandon:
Yes, I think so.  But I think we should leave a readme of the same name, that points to the docs, so that folks don't end up with broken links.  (And we should always do this, broken links are the worst in terms of UX.  But having a document that just points you to the new, actual documentation, is perfectly fine and refreshing.)

So two things:
1. Let's do this little redirect/description page back to the docs in this case.
2. Let's do this little redirect/description page in all other cases/situations.

Thoughts?

VanDavv:
Agreed @Luxonis-Brandon, I'll add these redirects in this PR and circle back

Luxonis-Brandon:
Thanks!

VanDavv:
@Luxonis-Brandon I've added info about migrated examples. Also added the placeholders for gaze-estimation and pedestrian reid, that were renamed in #69

Luxonis-Brandon:
Great thanks!

PR #41, opened 2021-01-18
Add gen2-nn-sync example

VanDavv:
This example shows how to present the neural network inference results on the inferenced frames.
Two ways of frame sync were used:
- Device sync, using DepthAI device-side queues (results of face detections are presented this way)
- Host sync, using built-in `queue` module  (results of landmarks detection are presented this way)


gen2-nn-sync/requirements.txt
```diff
@@ -0,0 +1,4 @@
+opencv-python==4.2.0.34; platform_machine != "armv7l"
```
szabi-luxonis, line 1: Use `opencv-python==4.5.1.48` for all platforms.
Same in other examples (if applicable)

szabi-luxonis (changes requested):

gen2-nn-sync/requirements.txt
VanDavv: Sure, changed and tested in this example and also updated others

szabi-luxonis (approved):
LGTM

PR #43, opened 2021-01-19
Add sequence number sync example

VanDavv:
This example demonstrates how to synchronize incoming frames using their sequence numbers. 
This allows displaying frames taken in exact same moment


VanDavv:
@themarpe is it possible now to make seq numbers from rgb camera to match the ones from mono cameras?
RGB starts earlier than Mono, and we had a sync mode in gen1 to tell the RGB camera to "wait" for mono cameras before 
increasing the sequence number value on it's frames.

gen2-seq-num-sync/requirements.txt
```diff
@@ -0,0 +1,5 @@
+opencv-python==4.5.1.48
+numpy==1.18.3
```
szabi-luxonis, line 2: numpy version doesn't match the one that comes with opencv 4.5. May cause issues.
`numpy==1.19.5`

szabi-luxonis (changes requested):

szabi-luxonis (approved):
LGTM

PR #45, opened 2021-01-22
open3d fix

saching13:
Modified installation of dependencies just like szabi did in depthai repository. And added note in the examples that make use of open3d and tested them.
Modified example folders:
- gen2-camera-demo - Tested
- gen2-spi - Not tested @jonngai  I tried it using 1092 but got some errors. can you check on that ?
- pcl-projection-rgb -  Tested
- point-cloud-projection - Tested


gen2-camera-demo/main.py
```diff
@@ -41,7 +41,10 @@
 pcl_converter = None
 if point_cloud:
     if out_rectified:
-        from projector_3d import PointCloudVisualizer
+        try:
+            from projector_3d import PointCloudVisualizer
+        except ImportError as e:
+            raise ImportError(f"\033[1;5;31mError occured when importing PCL projector: {e}. Try disabling the point cloud \033[0m ")
```
VanDavv, line 47: User cannot disable pcl since it's hardcoded as true, we need to add command line params also

saching13: Thanks. Added `-pcl` arg to disable it.

gen2-camera-demo/main.py
```diff
@@ -14,11 +15,17 @@
 But like on Gen1, either depth or disparity has valid data. TODO enable both.
 '''
 
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-pcl", "--pointcloud", help="display a square of a given number", default=False, action="store_true")
```
VanDavv, line 20: this help message should be a bit more meaningful (copypaste?)

gen2-camera-demo/main.py
saching13: Aahh right.

gen2-camera-demo/main.py
saching13: Oh. I actually copy pasted and noticed while testing and changed it. forgot to push I think will push along with fix for above error.

VanDavv (approved):
LGTM

PR #51, opened 2021-01-28
Refactor Gaze Estimation example

VanDavv:
Refactored and made it work faster, demo below

[Gaze Example Demo](https://user-images.githubusercontent.com/5244214/106155520-0f483d00-6181-11eb-8b95-a2cb73cc4bac.mp4)

### DIFF
--- a/gaze-estimation/README.md
+++ b/gaze-estimation/README.md
@@ -2,7 +2,7 @@
 
 This example demonstrates how to run 3 stage inference (3-series, 2-parallel) on DepthAI using Gen2 Pipeline Builder.
 
-[![Gaze Example Demo](https://user-images.githubusercontent.com/5244214/96713680-426c7a80-13a1-11eb-81e6-238e3decb7be.gif)](https://www.youtube.com/watch?v=OzgK5-APxBU)
+[![Gaze Example Demo](https://user-images.githubusercontent.com/5244214/106155937-4fa7bb00-6181-11eb-8c23-21abe12f7fe4.gif)](https://user-images.githubusercontent.com/5244214/106155520-0f483d00-6181-11eb-8b95-a2cb73cc4bac.mp4)
 
 
 Original OpenVINO demo, on which this example was made, is official [here](https://docs.openvinotoolkit.org/2021.1/omz_demos_gaze_estimation_demo_README.html) from Intel and implemented with the NCS2 with nice diagrams and explanation, [here](https://github.com/LCTyrell/Gaze_pointer_controller), from @LCTyrell.
--- a/gaze-estimation/main.py
+++ b/gaze-estimation/main.py
@@ -181,6 +181,7 @@ class Main:
         self.pose = None
         self.gaze = None
 
+        self.running = True
         self.fps = FPS()
         self.fps.start()
 
@@ -189,7 +190,7 @@ class Main:
         landmark_in = self.device.getInputQueue("landmark_in")
         pose_in = self.device.getInputQueue("pose_in")
 
-        while True:
+        while self.running:
             if self.frame is None:
                 continue
             bboxes = np.array(face_nn.get().getFirstLayerFp16())
@@ -216,7 +217,7 @@ class Main:
         pose_nn = self.device.getOutputQueue("pose_nn")
         gaze_in = self.device.getInputQueue("gaze_in")
 
-        while True:
+        while self.running:
             face_bbox = self.face_box_q.get()
             left = face_bbox[0]
             top = face_bbox[1]
@@ -240,7 +241,7 @@ class Main:
 
     def gaze_thread(self):
         gaze_nn = self.device.getOutputQueue("gaze_nn")
-        while True:
+        while self.running:
             self.gaze = np.array(gaze_nn.get().getFirstLayerFp16())
 
     def should_run(self):
@@ -253,9 +254,13 @@ class Main:
             return self.cap.read()
 
     def run(self):
-        threading.Thread(target=self.face_thread, daemon=True).start()
-        threading.Thread(target=self.land_pose_thread, daemon=True).start()
-        threading.Thread(target=self.gaze_thread, daemon=True).start()
+        self.threads = [
+            threading.Thread(target=self.face_thread, daemon=True),
+            threading.Thread(target=self.land_pose_thread, daemon=True),
+            threading.Thread(target=self.gaze_thread, daemon=True)
+        ]
+        for thread in self.threads:
+            thread.start()
 
         while self.should_run():
             read_correctly, self.frame = self.get_frame()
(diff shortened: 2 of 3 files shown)


Luxonis-Brandon:
Wow, looks great - thanks!

raymondlo84 (commented):
You may want to also review the requirement.txt or user will see this issue.

ModuleNotFoundError: No module named 'imutils'

raymondlo84:
File "main.py", line 312, in <module>
    Main().run()
  File "main.py", line 302, in run
    if cv2.waitKey(1) == ord('q'):
KeyboardInterrupt
Assertion failed: (pthread_mutexAssertion failed: (pthread_mutex_destroy(mutex) == 0), function _lock(mutex) == 0), function usbusbi_mutex_destroy, file ./os/thi_mutex_lock, file ./os/threads_reads_posix.h, line 58.
posix.h, line 46.
zsh: abort      python3 main.py -cam

... FYI.

VanDavv:
Thanks for the test @raymondlo84 ! 
I just pushed a fix with missing requirement as you've mentioned.
I will also implement a mechanism to exit the threads gracefully, but just out of curiosity - doesn't daemon threads do this already, as stated [here](https://docs.python.org/3/library/threading.html#threading.Thread.daemon)? `The entire Python program exits when no alive non-daemon threads are left` - so does this fix only solve the misleading console prints or is it somehow more beneficial?

raymondlo84:
Last and not least, I think the left and right eye bounding boxes should be scaled accordingly too? It seems now it's statics?

VanDavv:
@raymondlo84 yes, they are scaled statically because it's required by the neural network

raymondlo84:
> Thanks for the test @raymondlo84 !
> I just pushed a fix with missing requirement as you've mentioned.
> I will also implement a mechanism to exit the threads gracefully, but just out of curiosity - doesn't daemon threads do this already, as stated [here](https://docs.python.org/3/library/threading.html#threading.Thread.daemon)? `The entire Python program exits when no alive non-daemon threads are left` - so does this fix only solve the misleading console prints or is it somehow more beneficial?

Good question. =) I need to dig up documentation further as well. But the join is working and just tested it. Yes, the bounding box for left and right eye... Did we do it correctly?

raymondlo84:
> @raymondlo84 yes, they are scaled statically because it's required by the neural network

But the bounding box drawing is not reflecting that. I guess it's a pure UI issue?

Luxonis-Brandon:
Yes, it seems that way.  

So I'm thinking we should merge this (since it's faster) and then circle back to fix that as a separate effort.  Thoughts?

VanDavv:
@raymondlo84 This slight delay in bounding boxes is an issue with synchronization and we're working on this right now, so we'll circle back and improve this example further.

However, on our [discord channel](https://discord.com/channels/790680891252932659/791012940093849600/804629990490505236), user Boris provided a video where the example crashed, and I provided fixes for this issue [here](https://github.com/luxonis/depthai-experiments/pull/51/commits/3b8e6dc581ca029cae0e13345b3ec9a3bbda2d52).

@pamolloy @themarpe can you please take a look if you can see any issues with this refactor or some possible improvements to be made?

pamolloy (approved):

pamolloy:
Looks great! I did just notice the following when I exit:
```
python3: io.c:2115: handle_events: Assertion `ctx->pollfds_cnt >= internal_nfds' failed.
Aborted (core dumped)
```

VanDavv:
@pamolloy I have noticed similar issue when I'm unable to join the threads (a graceful thread exit I added last time didn't work), so then I just exit the app and use a fact that they're daemon threads and they have to exit. This message may come from DepthAI when it's terminated in a middle of an operation

themarpe:
The `Abort` issue stems from XLink and the way it operates with libusb. Something to be checked in the future when time allows.

themarpe (approved):
LGTM

raymondlo84:
> @pamolloy I have noticed similar issue when I'm unable to join the threads (a graceful thread exit I added last time didn't work), so then I just exit the app and use a fact that they're daemon threads and they have to exit. This message may come from DepthAI when it's terminated in a middle of an operation

Screen Shot 2021-01-29 at 11 03 45 AM

It got the same crash again, without the join thread. That's now very weird. =) 

Assertion failed: (pthread_mutexAssertion failed: (pthread_mutex_destroy(mutex) == 0), function _lock(mutex) == 0), function usbusbi_mutex_destroy, file ./os/thi_mutex_lock, file ./os/threads_reads_posix.h, line 58.

....

-ray

raymondlo84:
> @raymondlo84 This slight delay in bounding boxes is an issue with synchronization and we're working on this right now, so we'll circle back and improve this example further.
> 
> However, on our [discord channel](https://discord.com/channels/790680891252932659/791012940093849600/804629990490505236), user Boris provided a video where the example crashed, and I provided fixes for this issue [here](https://github.com/luxonis/depthai-experiments/pull/51/commits/3b8e6dc581ca029cae0e13345b3ec9a3bbda2d52).
> 
> @pamolloy @themarpe can you please take a look if you can see any issues with this refactor or some possible improvements to be made?

Oh, delay yes. But the bounding box (SIZE) for the eyes seem to be off and that throws me off a bit more.

I agree with Brandon we should open a new PR to track that. This thread should focus on thread parallelization and debugs.

Luxonis-Brandon:
Thanks both.  And also given that the other (current version) is so slow/laggy, and that this is generally better - we should merge this and then do issue tracking off of the merged version.

Otherwise many will likely come upon the other example, say `wow, this device sucks!`, but it in the drawer, and move along.

raymondlo84:
> Thanks both. And also given that the other (current version) is so slow/laggy, and that this is generally better - we should merge this and then do issue tracking off of the merged version.
> 
> Otherwise many will likely come upon the other example, say `wow, this device sucks!`, but it in the drawer, and move along.

=D Take the bullet and move on! but let's keep tracking these refinements. 

-ray

Luxonis-Brandon:
> but let's keep tracking these refinements.

Yes, for sure!  So to confirm, the third issue (aside from the left/right drawing and the latency) is the crash upon termination of the demo, right?  If so, I'll make an issue for that.

VanDavv:
@raymondlo84 the bounding boxes for eyes are generated on the UI, so they can be changed. However, they are now reflecting what is being sent to the neural network. It accepts eye images scaled to 60x60, and from landmarks neural network I receive a single eye point, so using `padded_point(..., padding=30, ...)` is producing a correct bounding box that I can send further.

This is also why this bounding box is sometimes too big / too small - as it doesn't take the face dimensions into account.

One fix for this, that would be fairly easy to do, is to somehow correlate the face dimensions with eye size, so that for given face width/height we can estimate the eyes width/height. This way, we could draw a better bounding box and the scaled image for nn would contain less non-eye parts, so it may also improve efficiency

PR #57, opened 2021-02-03
Fix threading issue

VanDavv:
This PR resolves https://github.com/luxonis/depthai-experiments/issues/54

Includes a FW fix of libusb, threads are not daemons anymore, are gracefully joined.
The device object is destroyed before joining the threads, which causes blocking calls of `get()` to throw RuntimeError. This is then caught and handled properly

[Tested on Linux](https://youtu.be/jPSHYufs8wY) - works, no system errors
[Tested on Mac](https://youtu.be/GfPDqfyvOlY) - works, no system errors nor dialog boxes

### DIFF
--- a/gaze-estimation/main.py
+++ b/gaze-estimation/main.py
@@ -1,6 +1,7 @@
 import argparse
 import queue
 import threading
+import time
 from pathlib import Path
 
 import cv2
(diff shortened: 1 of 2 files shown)


themarpe (approved):

Luxonis-Brandon:
Sweet!  Thanks, both!

PR #63, opened 2021-02-05
Add deeplabv3 person decoding for gen2

szabi-luxonis:
```
git clone https://github.com/luxonis/depthai-experiments.git
cd depthai-experiments
git checkout gen2_deeplabv3_person
cd gen2-deeplabv3_person
python3 deeplabv3_person_256.py
```

RGB packets are synchronized with NN packets, so the performance is limited by NN, that's why it's "slow".


Luxonis-Brandon (approved):
image

Works well for me.  And actually it seems the same update rate as Gen1, which is nice.  Nice to have the sync, too.

PR #64, opened 2021-02-08
Add pose estimation example

VanDavv:
# Gen2 Pose Estimation Example

This example demonstrates how to run [Human Pose Estimation Network](https://docs.openvinotoolkit.org/latest/omz_models_intel_human_pose_estimation_0001_description_human_pose_estimation_0001.html) using Gen2 Pipeline Builder.

## Demo

[Gen2 Age & Gender recognition](https://www.youtube.com/watch?v=Py3-dHQymko "Human pose estimation on DepthAI")

TODO:

- [ ] Improve inference performance

### DIFF
--- a/gen2-human-pose/README.md
+++ b/gen2-human-pose/README.md
@@ -1,0 +1,41 @@
+# Gen2 Pose Estimation Example
+
+This example demonstrates how to run [Human Pose Estimation Network](https://docs.openvinotoolkit.org/latest/omz_models_intel_human_pose_estimation_0001_description_human_pose_estimation_0001.html) using Gen2 Pipeline Builder.
+
+
+## Demo
+
+WIP
+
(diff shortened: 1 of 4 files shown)


VanDavv:
The code processing the results is very fragile, and some optimizations I had in mind ended up with a situation when changing one line of code broke another one. The main thing I had in mind was to use numpy array operations instead of loops - but methods like `getValidPairs` are for now too complex to dig in, the amount of work compared to potential gains is too high.

I managed to find some minor improvements, also it's possible to display the results on a full frame. I was also able to use process pools to make the calculations in parallel, but I didn't include it as there was no performance gain from these operations (time to sync between processes was similar to the time gained from parallelization).

I'll leave this example in the current state as it works stable and optimizations can be made further if there will be a need/requests from users

PR #68, opened 2021-02-15
Update FW w/ openvino 2021.2 support; update blob compiled for openvi…

szabi-luxonis:
…no 2021.2 from pinto zoo

Blob is taken and compiled from PINTO zoo.
Instructions of compilation in script description.

It's running at 30 fps vs 24-5 previously.

### DIFF
--- a/gen2-deeplabv3_person/deeplabv3_person_256.py
+++ b/gen2-deeplabv3_person/deeplabv3_person_256.py
@@ -15,6 +15,15 @@ python3 -m pip install -r requirements.txt
 python3 deeplabv3_person_256.py -cam rgb
 Possible input choices (-cam):
 'rgb', 'left', 'right'
+
+Blob taken from the great PINTO zoo
+
+git clone git@github.com:PINTO0309/PINTO_model_zoo.git
+cd PINTO_model_zoo/026_mobile-deeplabv3-plus/01_float32/
+./download.sh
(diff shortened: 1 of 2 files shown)


Luxonis-Brandon:
Thanks, this is great.  And love that we are using PINTO0309's models.  So I just tried running this and I am seeing the following on installing the requirements.txt:

```
python3 -m pip install requirements.txt
Defaulting to user installation because normal site-packages is not writeable
ERROR: Could not find a version that satisfies the requirement requirements.txt
ERROR: No matching distribution found for requirements.txt
```

On MacOS and commands I ran to check if this a PEBKAC error:
image

```
cd depthai-experiments 
git pull  
git checkout deeplabv3_improvements
cd gen2-deeplabv3_person 
git pull
python3 -m pip install requirements.txt
```

So I am wondering if the CI/CD failed on the build for macOS.  I think @alex-luxonis saw this on some of his builds recently.  CC: @themarpe .

Thoughts?

szabi-luxonis:
> python3 -m pip install requirements.txt

It's `python3 -m pip install **_-r_** requirements.txt`

Luxonis-Brandon:
Thanks @szabi-luxonis !  Man, I can't help myself on that one!  Testing again now!  (And sorry about the repeat error here.)

Luxonis-Brandon:
image
image

Yep, works great!  It's nice how fast and well-synched it is now (the previous version you did was well-synched too, but the Gen1 deeplab definitely had some offset).

Luxonis-Brandon (approved):

PR #72, opened 2021-02-24
Fix detection results in TF example

VanDavv:
Pedestrian Re-Identification

This PR introduces fixes to the TensorFlow usage example (posted [here](https://github.com/luxonis/depthai-experiments/issues/55#issuecomment-784497863)) that improved the inference results


Luxonis-Brandon (approved):
LGTM.  Since the previous one is broken I'm thinking we should merge this and then open separate issues if/when they exist.

PR #82, opened 2021-03-09
Update main.py

dhruvsheth-ai:
updated Lib to 2.0.0.1
Tested on armv7l

### DIFF
--- a/gen2-human-pose/main.py
+++ b/gen2-human-pose/main.py
@@ -167,11 +167,11 @@ with dai.Device(create_pipeline()) as device:
             if debug:
                 cv2.imshow("rgb", debug_frame)
 
-             key = cv2.waitKey(1)
-             if key == ord('q'):
+            key = cv2.waitKey(1)
+            if key == ord('q'):
                 break
                 
-             elif key == ord('t'):
+            elif key == ord('t'):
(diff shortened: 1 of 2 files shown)


Luxonis-Brandon:
Thanks!

gen2-human-pose/main.py
```diff
@@ -167,6 +169,12 @@ def get_frame():
 
             if cv2.waitKey(1) == ord('q'):
                 break
+            elif cv2.waitKey(2) == ord('t'):
```
szabi-luxonis, line 172: `cv2.waitKey(X) ` [waits](https://docs.opencv.org/2.4/modules/highgui/doc/user_interface.html#waitkey) for X milliseconds for keypress. 
Change to: 
```
key = cv2.waitKey(1)
if key == ord('q'):
     break
elif key == ord('t'):
```

szabi-luxonis (changes requested):

gen2-human-pose/main.py
szabi-luxonis: And thanks for the change. Otherwise LGTM

dhruvsheth-ai: Thanks, changed the file

szabi-luxonis (approved):
LGTM

PR #83, opened 2021-03-10
Gen2: Add wls filter

szabi-luxonis:
Run: `./install_dependencies.sh`
Then: `./main.py`


Luxonis-Brandon (approved):
Thanks for making this so quickly.  I tested it and I think this is good to merge.  

So two things I'm going to make as separate Github issues for later investigation:
1. Having the disparity confidence slider here would be great.  https://github.com/luxonis/depthai-experiments/issues/87
2. The disparity color range is really jumpy.  I'm guessing the range is normalized every frame or something like that.  So perhaps we could do a fixed normalization or some average over time as the scene changes maybe. https://github.com/luxonis/depthai-experiments/issues/86

I'll upload a video about `2`.  And on `1` I'm just super curious to see what some of the WLS looks like when the disparity confidence threshold is increased.

PR #88, opened 2021-03-10
Remove histogram equalization, scale disparity based on max range

### LINKED ISSUE #86: Investigations on Gen2 WLS Filter Example Disparity Color Range Jumpiness
In relation to https://github.com/luxonis/depthai-experiments/tree/master/gen2-wls-filter

The disparity color range is jumpy.  I don't know if this is out of my ignorance of how WLS works, or if it's something like the following. 

Below is a video showing it:
[https://youtu.be/IEqg0R2xA3g](https://youtu.be/IEqg0R2xA3g)

With that said, I'm guessing the color range is normalized every frame to the present min and max disparities, or something like that. So perhaps we could do a fixed normalization or some average over time as the scene changes maybe.

szabi-luxonis:
Regarding comments on https://github.com/luxonis/depthai-experiments/pull/83
Fixes #86 
Removed histogram equalization, scaled disparity based on max range (95)

There are still fluctuations, coming from WLS filter which can be seen on raw filtered disparity image.

Depth colorization removed since it doesn't makes sense when we have disparity on hands.

### DIFF
--- a/gen2-wls-filter/main.py
+++ b/gen2-wls-filter/main.py
@@ -66,19 +66,16 @@ class wlsFilter:
         self.lambdaTrackbar = trackbar('Lambda', self.wlsStream, 0, 255, 80, self.on_trackbar_change_lambda)
         self.sigmaTrackbar  = trackbar('Sigma',  self.wlsStream, 0, 100, 15, self.on_trackbar_change_sigma)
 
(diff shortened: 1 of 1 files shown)


Luxonis-Brandon (approved):
image
Looks great to me.  Uploading video as well to show the jumpiness is gone.

PR #90, opened 2021-03-11
Fix and improve human pose estimation example

VanDavv:
This PR improves human pose estimation example

[Gen2 Age & Gender recognition](https://www.youtube.com/watch?v=1dp2wJ_OqxI "Human pose estimation on DepthAI")

### DIFF
--- a/gen2-human-pose/README.md
+++ b/gen2-human-pose/README.md
@@ -5,8 +5,12 @@ This example demonstrates how to run [Human Pose Estimation Network](https://doc
 
 ## Demo
 
+### Camera
 [![Gen2 Age & Gender recognition](https://user-images.githubusercontent.com/5244214/107493701-35f97100-6b8e-11eb-8b13-02a7a8dbec21.gif)](https://www.youtube.com/watch?v=Py3-dHQymko "Human pose estimation on DepthAI")
 
+### Video file
+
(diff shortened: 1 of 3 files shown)


szabi-luxonis (changes requested):
Rename blob to match naming convention: https://artifacts.luxonis.com/artifactory/luxonis-depthai-data-local/network/

E.g.
`human-pose-estimation-0001_openvino_2021.2_6shave.blob `

Blob should be compiled for openvino 2021.2, and for 6/8 shaves with 2 inference threads.

VanDavv:
@szabi-luxonis I renamed them - I used 2021.1 to compile - can I leave them as is?

szabi-luxonis:
> @szabi-luxonis I renamed them - I used 2021.1 to compile - should they be changed to 2021.2 or can be left?

We should use the latest version in examples IMO.

Luxonis-Brandon:
Yes and if we use latest version, this has advantage the if folks are combining examples they all have the same OpenVINO version.

For OpenVINO versions we should try to like `jump` together across all examples if possible when we choose a new version.  That way the examples can be plug/played together without having differing versions of OpenVINO compilation which would cause them not to work and user confusion.

Thoughts?

VanDavv:
I updated the models to use 2021.2 OpenVINO version. To accomplish it, I released an update for BlobConverter to support 2021.2 conversion online - http://69.164.214.171:8084/

Luxonis-Brandon:
Thanks!

themarpe (approved):
LGTM

PR #92, opened 2021-03-12
Vehicle License Plate identification Example

wasauce:
Let me note first and foremost this PR is NOT ready to merge.

---

This is an example for doing vehicle license plate identification.

The example borrows code structure from: https://github.com/luxonis/depthai-experiments/tree/master/gen2-pedestrian-reidentification

The key reason being that this would allow me to pass in an example input video -- and not just do live camera capture.

Current problems:
1/ On my MacBook Pro -- after ~ 10-15 seconds the demo script hangs and I haven't yet determined the cause. Surely it is something "dumb" on my behalf

2/ I am not yet actually extracting the license plate content. I should add something like `license-plate-recognition-barrier-0001` and `vehicle-attributes-recognition-barrier-0039` to extract more information

Note to self -- see https://github.com/luxonis/depthai-experiments/issues/47


Luxonis-Brandon:
Thanks for putting this together!  And CC: @VanDavv as he will be helping on this as well soon.

VanDavv:
Hi @wasauce !
Please check out https://github.com/luxonis/depthai-experiments/pull/94 - I've added a detection + recognition neural networks (so it's a 2 stange nn example).
It doesn't have `vehicle-attributes-recognition-barrier-0039` added yet, but I've added a note to add it so it's not forgotten

PR #94, opened 2021-03-14
Gen2 license plate example

VanDavv:
# Gen2 License Plates Recognition

This example demonstrates how to run 2 stage inference on DepthAI using Gen2 Pipeline Builder.
First, a license plate is detected on the image and then the cropped license frame is sent to text detection network, 
which tries to decode the license plates texts

> :warning: **This demo is adjusted to detect and recognize chinese license plates**! It may not work with other license plates

## Demo

[Gen2 License Plates recognition](https://www.youtube.com/watch?v=tB_-mVVNIro "License Plates recognition on DepthAI")


VanDavv:
This PR is now ready to merge IMO.
It has license plate detection + recognition + car attribute estimator.

I've noticed though that `vehicle-license-plate-detection-barrier-0106` has a poor performance in terms of detecting the cars - so I've added a new network - `vehicle_detection_adas_0002` - that is detecting the cars way better and is later on feeding these results to car attributes network.
This however decreases the overall performance of the example, that's why I didn't add it to this PR and pushed to a separate branch instead - [`gen2_license_plate_example_separate_veh_nn`](https://github.com/luxonis/depthai-experiments/tree/gen2_license_plate_example_separate_veh_nn)

VanDavv:
Thanks @wasauce , just added the mention

Luxonis-Brandon:
@VanDavv - can we default-use the license plate detector trained via reoboflow?  And then mention that it was trained via roboflow, and that others can retrain for their region/state/nation using roboflow?

wasauce:
@Luxonis-Brandon ooh - the Roboflow one would be great -- so we can better understand how to use Roboflow to build a model :) I have a large collection of US license plate photos from http://vision.ucsd.edu/belongie-grp/research/carRec/car_data.html and I am interested in building a model via RoboFlow

wasauce:
@VanDavv - I tried this branch and I tried `gen2_license_plate_example_separate_veh_nn`

I think you should bring in all the changes/enchancements from `gen2_license_plate_example_separate_veh_nn` - the performance degradation seemed sufficient and frankly not that noticeable. 

A question -- in `gen2_license_plate_example_separate_veh_nn` - the attributes of the vehicles continued to be identified -- whereas this branch the attributes got "Stuck" after the white car and then stopped updating. I would use this as a second argument for why `gen2_license_plate_example_separate_veh_nn` is good!

VanDavv:
Thanks @wasauce for the feedback, I agree and will merge these changes here

> A question -- in gen2_license_plate_example_separate_veh_nn - the attributes of the vehicles continued to be identified -- whereas this branch the attributes got "Stuck" after the white car and then stopped updating

This is the issue I tried to solve with this additional branch. The `vehicle-license-plate-detection-barrier-0106` is adjusted to the barrier scenario, so I guess it learned to detect the cars looking top to bottom, not facing directly - that's why it just doesn't produce confident-enough results and is filtered out by the confidence threshold.
I tried to lower the confidence threshold for the cars but at some point it started to get to noisy.

themarpe (changes requested):
Rename `chineese_traffic.mp4` to `chinese_traffic.mp4`
Otherwise LGTM

wasauce:
@Luxonis-Brandon - do you have any information you can share about what you wrote above: "default-use the license plate detector trained via roboflow" or information/documentation about how to roboflow and export a model for use with DepthAI?

PR #96, opened 2021-03-15
Migrate examples to use depthai 2.0.0.1 and follow a common blob naming format

VanDavv:
While migrating examples to the latest PyPi release, and making sure the blobs are using the latest OpenVINO and have correct blob names, I found two examples that stopped working as they used to:

- Pedestrian Reidentification network (`person-reidentification-retail-0031`) is no longer available in 2021.2, was removed in 2020.4.  The new network that I tried to use - [person_reidentification_retail_0288](https://docs.openvinotoolkit.org/latest/omz_models_intel_person_reidentification_retail_0288_description_person_reidentification_retail_0288.html) and [person_reidentification_retail_0277](https://docs.openvinotoolkit.org/latest/omz_models_intel_person_reidentification_retail_0277_description_person_reidentification_retail_0277.html) - accept an input image in higher resolution (now `128x256`, previously `48x96`), which may be a cause of a significant inference speed drop when these networks start to process the frames. 

  Code for this example is [here](https://github.com/luxonis/depthai-experiments/tree/pedestrian_reid_2021_2) as was not included in this PR due to the performance issues

- Tensorflow Image Classification example has accuracy issues when model is changed to the one compiled for 6 or 8 shaves. Where previously is recognized a dandelion (like on a demo image) it does no more. That's why this example is also not included in this PR, and its source code is [here](https://github.com/luxonis/depthai-experiments/tree/tf-image-classification-2021.2)

Other than these two, the rest of the examples work fine.


szabi-luxonis:
You can use openvino IR model compiled for 2020.1/3/4 for `person-reidentification-retail-0031` and use myriad compiler from 2021.2, it will likely work. That's how I compiled deeplabv3 for openvino 2021.2 since I had only 2020.1 IR model from gen1. @VanDavv

VanDavv:
That's very clever, will do. Thanks @szabi-luxonis !

szabi-luxonis (approved):
LGTM

PR #97, opened 2021-03-16
fix (pose estimation) compiled blob with 2021.2 openvno version

Erol444:
as per ticket: https://app.asana.com/0/1199908639261369/1200022744474313/f


Erol444:
Already done: https://github.com/luxonis/depthai-experiments/pull/90

PR #104, opened 2021-03-31
Migrate collecting training data to Gen2

VanDavv:
This PR migrates data collecting experiment to use Gen2.

Downside, compared to gen1, is the timestamp sync between rgb and mono frames, because we cannot sync the rgb and mono sequence numbers (yet) in Gen2

Upside, compared to gen1, is the ability to set exposure and manual focus of the RGB camera dynamically


Luxonis-Brandon:
Thanks.  And I think @alex-luxonis may have just gotten timesync between RGB and grayscale in Gen2 working here: https://github.com/luxonis/depthai-python/pull/147/commits/653ba10b9a3865ee3ed2dc6593e4888b5c5fe03a

He can confirm though - as I'm not 100% sure.

Erol444 (approved):
Just curious, would it be useful for users a feature to specify FPS in args? Maybe for some cases where changes are gradual and this would save disk space.
Otherwise LGTM

cafemoloko (approved):

PR #107, opened 2021-04-08
Fixed the video preview size and also improved the result fonts

raymondlo84:
This is mostly the usability fix. The old code was using 180x180 as a preview and it's very hard to see the results.


Luxonis-Brandon:
Thanks @raymondlo84 !  Good idea to make this a lot clearer when looking at it!  And appreciate the PR here!

Erol444 (approved):
Thanks!

VanDavv (approved):
Just tested and it works great, thanks @raymondlo84 ! (and sorry for the delay)

PR #110, opened 2021-04-19
migrated the `triangulation-3D-visualizer` to gen2. 

Erol444:
Used Script node to crop the image


Luxonis-Brandon:
Sweet.  This will be great for many to have.  And I love that it uses the script node as that affords the capability for folks to customize it!

gen2-triangulation/README.md
```diff
@@ -0,0 +1,39 @@
...
+(i.e. could be done on an ESP32), we leave the triangulation itself to the host.
+
+This 3D visualizer is for the facial landmarks demo, and uses OpenGL and OpenCV.
+Consider it a draft/reference at this point.
+
+## Demo
+
+[![Spatial AI](https://user-images.githubusercontent.com/5244214/90748450-7c9e9a80-e2d2-11ea-9e9e-da65b5d9e6f0.gif)](https://www.youtube.com/watch?v=Cs8xo3mPBMg "3D Facial Landmark visualization")
+
+Note that this is Gen1 video demo
```
VanDavv, line 15: I think we should record a new video instead

VanDavv (changes requested):
Can we remove the old `triangulation-3D-visualizer` (and leave readme note that it was moved to gen2, like [here](https://github.com/luxonis/depthai-experiments/tree/master/gaze-estimation)).

This way, we should have less and less codebase written in Gen1, so we won't have to maintain it and will direct users to use Gen2 by default

Luxonis-Brandon:
> removed gen1 triangulation experiment

You may have already done this, but please make sure there is a redirect in place so the old link still works.  As we want to allow folks to link to any of our examples, experiments, anything with confidence that their customers or fans (say if it's a Youtube channel) doesn't end up with a link that goes nowhere.

Erol444:
Added the video
giff

gen2-triangulation/README.md
Erol444: added now

Luxonis-Brandon:
Great demo - thanks!

VanDavv (approved):
LGTM

PR #111, opened 2021-04-22
Get RGB frame working with point cloud.

mribbons:


Luxonis-Brandon:
Thanks @mribbons!  CC: @luxonis-sachin

Luxonis-Brandon:
Thanks @mribbons!  CC: @luxonis-sachin

saching13:
Since there are better improvements in #112 can we close this ?

mribbons:
@saching13 The only reason I made a different branch and PR for #112 was that it requires the library versions as per luxonis:camera_demo_rgb_stereo. The code in my #111 will run against master... But I guess the recipient won't have access to this change anyway. Happy to close this one.

PR #112, opened 2021-04-22
Camera demo rgb stereo align

mribbons:
Resolve RGB alignment issue by using stereo.setDepthAlign(dai.CameraBoardSocket.RGB)

Note that we need the python libs as per https://github.com/luxonis/depthai-python/pull/147/commits/0a27ebcf5301ba9166d014bbbb543e431b5af212 for this to work.

We get some warnings, and the disparity frames seem to flick back to front a bit.

If we run without -pcl, to test if it's a performance issue on host side,
the left/right flickering is quite fast.

python main.py -ext 1 -sub 0 -pcl -rgb -lr 0

[14442C1091EB8FD000] [8.681] [StereoDepth(5)] [warning] Median filter usage with LR-check/Extended/Subpixel is not yet implemented. The filter will be disabled
[14442C1091EB8FD000] [8.681] [StereoDepth(5)] [warning] DepthAlign option ignored, as it requires LR-check mode to be enabled

### DIFF
--- a/gen2-camera-demo/main.py
+++ b/gen2-camera-demo/main.py
@@ -393,21 +393,9 @@ def test_pipeline():
                     continue
                 #print("Received frame:", name, "tstamp", image.getTimestamp().total_seconds())
                 # Skip some streams for now, to reduce CPU load
-                if name in ['left', 'right', 'rectified_left', 'rectified_right']: continue
+                if name in ['left', 'right']: continue
                 frame = convert_to_cv2_frame(name, image)
                 cv2.imshow(name, frame)
-                # This code shows that stereo.setDepthAlign(dai.CameraBoardSocket.RGB) is working
-                # if name == 'disparity' and last_frame_rgb_video is not None:
-                #     rgb = last_frame_rgb_video
-                #     if len(rgb) != 720:
-                #         rgb = cv2.resize(rgb, (1280,720))
-                        
-                #     #print('shape grey: {}, {}'.format(frame.shape, rgb.shape))
-                #     if len(frame.shape) < 3:                        
-                #         blended = cv2.addWeighted(rgb, 0.5, cv2.cvtColor(frame, cv2.COLOR_GRAY2BGR), 0.5, 0)
-                #     else:
-                #         blended = cv2.addWeighted(rgb, 0.5, frame, 0.5, 0)
-                #     cv2.imshow('rectified_blend', blended)
             if cv2.waitKey(1) == ord('q'):
                 break
 
--- a/gen2-camera-demo/main.py
+++ b/gen2-camera-demo/main.py
@@ -253,6 +253,7 @@ def build_pipeline(pipeline):
 # The operations done here seem very CPU-intensive, TODO
 def convert_to_cv2_frame(name, image):
     global last_rectif_right
+    global last_frame_rgb_video
     baseline = 75  # mm
     focal = right_intrinsic[0][0]
     max_disp = 96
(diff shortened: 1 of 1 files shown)


Luxonis-Brandon:
Awesome, thanks!  CC: @saching13

gen2-camera-demo/main.py
```diff
@@ -372,9 +393,21 @@ def test_pipeline():
...
+                #     rgb = last_frame_rgb_video
+                #     if len(rgb) != 720:
+                #         rgb = cv2.resize(rgb, (1280,720))
+                        
+                #     #print('shape grey: {}, {}'.format(frame.shape, rgb.shape))
+                #     if len(frame.shape) < 3:                        
+                #         blended = cv2.addWeighted(rgb, 0.5, cv2.cvtColor(frame, cv2.COLOR_GRAY2BGR), 0.5, 0)
+                #     else:
+                #         blended = cv2.addWeighted(rgb, 0.5, frame, 0.5, 0)
+                #     cv2.imshow('rectified_blend', blended)
```
saching13, line 410: Can you remove this commented code.

gen2-camera-demo/main.py
```diff
@@ -372,9 +393,21 @@ def test_pipeline():
                     continue
                 #print("Received frame:", name, "tstamp", image.getTimestamp().total_seconds())
                 # Skip some streams for now, to reduce CPU load
-                if name in ['left', 'right']: continue
+                if name in ['left', 'right', 'rectified_left', 'rectified_right']: continue
```
saching13, line 396: I think it is better to allow `rectified_left`, `rectified_right` so that users can take a look at them and see how much they are rectified.

saching13:
Thanks alot for the PR.

gen2-camera-demo/main.py
mribbons: This is done.

gen2-camera-demo/main.py
mribbons: I did mean to put this back, was testing performance.
This is done @saching13

gen2-camera-demo/main.py
```diff
@@ -173,13 +175,16 @@ def build_pipeline(pipeline):
...
         stereo.setRectifyEdgeFillColor(0)  # Black, to better see the cutout
         stereo.setMedianFilter(median)  # KERNEL_7x7 default
         stereo.setLeftRightCheck(lrcheck)
         stereo.setExtendedDisparity(extended)
         stereo.setSubpixel(subpixel)
-        stereo.setDepthAlign(dai.StereoDepthProperties.DepthAlign.CENTER)
+        if enable_rgb:
+            stereo.setDepthAlign(dai.CameraBoardSocket.RGB)
+        else:
+            stereo.setDepthAlign(dai.StereoDepthProperties.DepthAlign.CENTER)
```
saching13, line 187: Can you make else/default alignment to be in rectified left camera?  
on **WHY** : if the else statement is center aligned. The pcl projection when non-rgb is using rectified right camera frame and doesn't align well.
Maybe we can add another argument to move it to the center and have PCL projection for it without any stream overlap.

Thoughts?

saching13: sounds good. Make default to `dai.StereoDepthProperties.DepthAlign.RECTIFIED_RIGHT` instead. 
Thanks a lot again for the additional changes.

mribbons: That is done. Note that I have defaulted to RGB when -rgb is enabled but --stereo-align isn't specifed and added a warning otherwise.

eg
```
Warning: --stereo_align should be set to rgb when -rgb is enabled.
```

Erol444:
Closing due to inactivity.

PR #113, opened 2021-04-25
Gen2 changes

raudette:
I made changes to support Gen2.  They are significant, and mostly taken from the 08 - RGB & MobilenetSSD example.
I found the MJPEG streaming didn't work with Firefox or Safari - it works fine with Chrome, and an iOS app called IPCams.

### DIFF
--- a/mjpeg-streaming/README.md
+++ b/mjpeg-streaming/README.md
@@ -25,7 +25,7 @@ Run the application
 python3 main.py
 ```
 
-To see the streamed frames, open [localhost:8090](http://localhost:8090)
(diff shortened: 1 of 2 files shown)


Luxonis-Brandon:
Thank you!

Erol444:
This is great @raudette! Two minor changes that would be neat:
- Move the experiment into a new folder `gen2-mjpeg-streaming` and make and redirect users to it ([like here](https://github.com/luxonis/depthai-experiments/tree/master/coronamask))
- It would be really nice to add a `print("DepthAI is up&running. Navigate to 'localhost:8090' with Chrome to see the mjpeg stream")` after device initialization

Thanks a ton!

PR #114, opened 2021-04-28
added deeplab + depth cropping experiment

Erol444:
deeplabGif

### DIFF
--- a/gen2-deeplabv3_depth/main.py
+++ b/gen2-deeplabv3_depth/main.py
@@ -20,15 +20,12 @@ python3 /opt/intel/openvino/deployment_tools/model_optimizer/mo_tf.py   --input_
 '''
 
 parser = argparse.ArgumentParser()
-parser.add_argument("-nn", "--nn_model", help="select camera input source for inference", default='models/deeplab_v3_plus_mvn2_decoder_256_openvino_2021.2_6shave.blob', type=str)
+parser.add_argument("-shape", "--nn_shape", help="select NN model shape", default=256, type=int)
+parser.add_argument("-nn", "--nn_path", help="select model path for inference", default='models/deeplab_v3_plus_mvn2_decoder_256_openvino_2021.2_6shave.blob', type=str)
 args = parser.parse_args()
 
-nn_path = args.nn_model
-
-
-nn_shape = 256
-if '513' in nn_path:
-    nn_shape = 513
+nn_shape = args.nn_shape
+nn_path = args.nn_path
 
 def decode_deeplabv3p(output_tensor):
     class_colors = [[0,0,0],  [0,255,0]]
--- a/gen2-deeplabv3_depth/main.py
+++ b/gen2-deeplabv3_depth/main.py
@@ -38,6 +38,11 @@ def decode_deeplabv3p(output_tensor):
 def show_deeplabv3p(output_colors, frame):
     return cv2.addWeighted(frame,1, output_colors,0.2,0)
 
+def dispay_colored_depth(frame, name):
+    frame_colored = cv2.normalize(frame, None, 255, 0, cv2.NORM_INF, cv2.CV_8UC1)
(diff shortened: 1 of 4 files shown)


szabi-luxonis (changes requested):
Depth frame is mapped to rectified right output, not right, so for perfect alignment rectified right should be used as NN input

gen2-deeplabv3_depth/main.py
```diff
@@ -0,0 +1,167 @@
...
+'''
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-nn", "--nn_model", help="select camera input source for inference", default='models/deeplab_v3_plus_mvn2_decoder_256_openvino_2021.2_6shave.blob', type=str)
+args = parser.parse_args()
+
+nn_path = args.nn_model
+
+
+nn_shape = 256
```
VanDavv, line 29: let's make an argument `nn_shape` and make it default to `256` - checking for string in path can cause some unclear edge-cases (e.g. when user will provide an absolute path and have `513` in their username)

gen2-deeplabv3_depth/main.py
```diff
@@ -0,0 +1,167 @@
...
+        if in_right is not None:
+            cv2.imshow("right", in_right.getCvFrame())
+
+        if in_cropped is not None:
+            frame = in_cropped.getCvFrame()
+
+        if in_depth is not None:
+            depth_frame = in_depth.getFrame()
+            # Crop the depth map
+            depth_frame = depth_frame[0:400, 120:520]
```
VanDavv, line 128: can we make it dynamic? I think the first `0:400` can be replaced with `:` (to take all), maybe we could make `120:520` not a hard-coded value too

gen2-deeplabv3_depth/main.py
```diff
@@ -0,0 +1,167 @@
...
+            # Resize it so it will match the NN output
+            depth_frame = cv2.resize(depth_frame, (nn_shape,nn_shape))
+
+            # Remove this to disable showing depth frames
+            depthFrameColor = cv2.normalize(depth_frame, None, 255, 0, cv2.NORM_INF, cv2.CV_8UC1)
```
VanDavv, line 133:
```suggestion
            depth_frame = cv2.normalize(depth_frame, None, 255, 0, cv2.NORM_INF, cv2.CV_8UC1)
```
if we would rename this to depth_frame, then in line 161 we wouldn't have to do the same operations again

VanDavv (changes requested):
This example is great to have! Noticed some minor issues but I'm very excited to add this to our experiments page!

gen2-deeplabv3_depth/main.py
VanDavv: it can be also `depthFrame` (renamed in a whole file) to match the camelCase standard

gen2-deeplabv3_depth/main.py
Erol444: or use imageManip node on device? Might be clearer for the users

gen2-deeplabv3_depth/main.py
Erol444: wanted to make it easy for users to add their custom code. I added a function that will color&display depth frames, so this piece of code isn't duplicated

gen2-deeplabv3_depth/main.py
Erol444: used imageManip to crop the frame (so it's 1:1 w:h) and quickly explained maths in the code

VanDavv (approved):
LGTM

Luxonis-Brandon:
Looks awesome - thanks!  Once we have the depth-RGB alignment released (https://github.com/luxonis/depthai-python/pull/147), we should update this to allow running off of right or RGB (and when run off RGB, we enable the depth-RGB alignment enabled.

PR #115, opened 2021-04-28
added gen2 people tracker

Erol444:
I will also add MCU side code that will be able to receive the tracklets, do a crude decode (same as on the host) and publish it to azure/aws. 
Demo:
gen2-people-counter


Luxonis-Brandon:
Looks great to me.

VanDavv (approved):

PR #116, opened 2021-05-01
Collision Avoidance minor change

dhruvsheth-ai:
Initially, the main Frame and the bird eye view would flicker a lot on Raspberry Pi, due to the same "Frame" name with the other `moveWindow` frame. Updated it and a minor change.

### DIFF
--- a/collision-avoidance/main.py
+++ b/collision-avoidance/main.py
@@ -50,7 +50,7 @@ class MainDebug(Main):
     def make_bird_frame(self):
         fov = 68.7938
         min_distance = 0.827
-        frame = np.zeros((300, 100, 3), np.uint8)
+        frame = np.zeros((384, 100, 3), np.uint8)
         min_y = int((1 - (min_distance - self.min_z) / (self.max_z - self.min_z)) * frame.shape[0])
         cv2.rectangle(frame, (0, min_y), (frame.shape[1], frame.shape[0]), (70, 70, 70), -1)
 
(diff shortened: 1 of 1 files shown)


Luxonis-Brandon:
Thanks!

This gave me an idea: we should get the full pose of the OAK-D from the IMU that’s onboard and use that to place the birds eye correctly even when the OAK-D is tilted relative to the earth.

dhruvsheth-ai:
> Thanks!
> 
> This gave me an idea: we should get the full pose of the OAK-D from the IMU that’s onboard and use that to place the birds eye correctly even when the OAK-D is tilted relative to the earth.

That is an interesting idea! I checked the demo again to understand it better. So, when it is at an angle with respect to the earth, the bird eye view, gives the distance in terms on distance from the angled oak-d. So something like, using the IMU to get the distance when the detected object is parallel to the surface? I have not tested the IMU example yet. I'll test it today to find whether it gives apparent coordinates, or relative.

Luxonis-Brandon:
Yes.  It might not be as relevant here as with the social distancing example.  But here's the point:

The camera has the XYZ distance of an object relative to how it is oriented.  So imagine this is on a bike, and the bike is leaned over at 30 degrees.  So this means the X-plane of the camera is now 30 degrees from the X-plane of the earth.  So They XZ measurements as shown from the birds-eye are off by appropriate trig between earth-x and camera-x axis.  Does that make sense?

dhruvsheth-ai:
That's a good catch. X-Y plane should be measured parallel from Earth in the Bird-eye view, while it is measured relative. So with the IMU example, the actual distance could be calculated.

This can be a used as reference - 
cropped_image

Since IMU is in gen2, the collision avoidance may have to be updated to gen2 as well. I've not done a gen1 to gen2 before, so I'll try to give it a shot to change the version.

Luxonis-Brandon:
Thanks.  Exactly.

VanDavv (approved):

VanDavv:
Thanks @dhruvsheth-ai , tested it and it works great

PR #125, opened 2021-05-08
gen2 record/replay

Erol444:
**This functionality (both recording and replaying) will be moved to `depthai/depthai_demo.py` script in the future.**

-------------------------

added record.py/replay.py scripts. Also added */data/* to gitignore.

TODO:
- Optionally save depth as well and when replaying, use saved depth instead of mono frames & stereo node
- Add readme
- Add demo video/gif
-

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .DS_Store
 __pycache__
+*/data/*
--- a/gen2-replay/README.md
+++ b/gen2-replay/README.md
@@ -28,10 +28,12 @@ optional arguments:
                         Set manual exposure time of the RGB camera [1..33000]
   -ei EXPOSURE_ISO, --exposure-iso EXPOSURE_ISO
                         Set manual exposure ISO of the RGB camera [100..1600]
-   -e, --encode         Encode mono frames into jpeg
-   -disp, --disp      Store disparity frames as well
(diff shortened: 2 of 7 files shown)


Luxonis-Brandon:
This will be so nice to have in depthai_demo.py.  And I really like the idea that was brought up internally of adding some kind of "dataset sharing" functionality, so community (and we) could prepare some common use case datasets (like driving) and whole community could easily use it by just specifying the name into the depthai_demo argument.  That will be so awesome for folks to be able to replicate work.

dhruvsheth-ai (commented):
https://github.com/luxonis/depthai-experiments/blob/gen2-replay/gen2-replay/README.md#replay-usage

Erol444:
> https://github.com/luxonis/depthai-experiments/blob/gen2-replay/gen2-replay/README.md#replay-usage
> 
> Just a suggestion, the -p comment could be changed to `Path of the stored data` to avoid confusion with the record.py path comment. :)
> 
> This script is really useful btw!!

you are totally right, we will keep that in mind when migrating this functionality into depthai_demo, thanks :)

Erol444:
a (very) late update: we have merged this functionality in a separate PR: https://github.com/luxonis/depthai-experiments/pull/202

PR #127, opened 2021-05-13
update OCR example to fix "0" detection problem

VanDavv:
One of our users reported an issue with detecting "0" (number) using the OCR example. I was able to reproduce it below

Screenshot from 2021-05-13 11-54-03

I analyzed the decoding code and found a check `t[i] == 0` that excluded the detection, but it was incorrect. The fill value for empty detections is in fact `36`, which resolves to `#` character after decoding - which we exclude in the next line.

With this PR, the detection of "0" (number) works correctly
Screenshot from 2021-05-13 11-48-16

### DIFF
--- a/gen2-ocr/main.py
+++ b/gen2-ocr/main.py
@@ -8,6 +8,7 @@ import depthai as dai
 import east
 
 pipeline = dai.Pipeline()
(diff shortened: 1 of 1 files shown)


Erol444 (approved):
thanks for finding this:)

PR #131, opened 2021-05-21
Add EfficientNet Image Classification Example with 1000 labels

dhruvsheth-ai:
Added the EfficentNet-b0 model Image Classification demo ffrom Open Model ZOO - https://github.com/openvinotoolkit/open_model_zoo/blob/master/models/public/efficientnet-b0/efficientnet-b0.md

This example contains 1000 labels which can be found at `classes.py`. Demo output -
2021-05-21-215934_848x480_scrot

result

Will be adding FPS to the script. Any other suggestions/changes are highly appreciated. If there can be made changes to the script to add NN optimisation as in here - https://github.com/luxonis/depthai-experiments/blob/master/gen2-efficientDet/README.md

That would be amazing too!

PS - Since this is a classification model, the FPS is similar to RGB so didn't add. If required, you'll can suggest some changes.


dhruvsheth-ai:
PS - It mostly classifies all common objects except human face. Really not aware about that bias.

Luxonis-Brandon:
Sweet this is great!  Thanks for making this!

gen2-efficentnet-classification/main.py
```diff
@@ -0,0 +1,152 @@
...
+# host, but you could also use ImageManip node to do it on device
+
+# Link video in with the detection network
+    
+if not args.camera and not args.video:
+    raise RuntimeError("No source selected. Please use either \"-cam\" to use RGB camera as a source or \"-vid <path>\" to run on video")
+
+debug = not args.no_debug
+camera = not args.video
+class_names = class_names()
```
VanDavv, line 29:
```suggestion
labels = class_names()
```
This line shadowed the imported `class_names` function

gen2-efficentnet-classification/main.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def to_planar(arr: np.ndarray, shape: tuple) -> np.ndarray:
+        resized = cv2.resize(arr, shape)
+        return resized.transpose(2, 0, 1)
+
+
+# Pipeline defined, now the device is assigned and pipeline is started
+with dai.Device(pipeline) as device:
+    device.startPipeline()
```
VanDavv, line 87: you can remove this line
```
DeprecationWarning: Device(pipeline) starts the pipeline automatically. Use Device() and startPipeline(pipeline) otherwise
```

VanDavv (approved):
I have really just two minor comments, but other than that LGTM, thanks @dhruvsheth-ai !

gen2-efficentnet-classification/main.py
dhruvsheth-ai: Sure! I have changed the script as per suggestion :)

PR #132, opened 2021-05-21
Add gen2-mono-raw-controls

alex-luxonis:
Displays both the `raw` output of a mono camera (left) and the ISP processed `out`.
(`raw` output added in https://github.com/luxonis/depthai-core/pull/135)

Key controls:
```
'IOKL' for manual exposure/iso:
  Control:      key[dec/inc]  min..max
  exposure time:     I   O     20..33000 [us]
  sensitivity iso:   K   L    100..1600
To go back to auto controls:
  'E' - autoexposure
Other camera controls:
'1' - AWB lock (true / false)
'2' - AE lock (true / false)
'3' - Select control: AWB mode
'4' - Select control: AE compensation
'5' - Select control: anti-banding/flicker mode
'6' - Select control: effect mode
'7' - Select control: brightness
'8' - Select control: contrast
'9' - Select control: saturation
'0' - Select control: sharpness
'[' - Select control: luma denoise
For the 'Select control: ...' options, use these keys to modify the value:
  '-' or '_' to decrease
  '+' or '=' to increase
'C' - capture a set of frames (PNG and unprocessed)
'Q' - quit
```

### DIFF
--- a/gen2-mono-raw-controls/main.py
+++ b/gen2-mono-raw-controls/main.py
@@ -1,0 +1,293 @@
+#!/usr/bin/env python3
+
+''' Control keys:
+'IOKL' for manual exposure/iso:
+  Control:      key[dec/inc]  min..max
+  exposure time:     I   O     20..33000 [us]
+  sensitivity iso:   K   L    100..1600
+To go back to auto controls:
+  'E' - autoexposure
+Other camera controls:
+'1' - AWB lock (true / false)
+'2' - AE lock (true / false)
+'3' - Select control: AWB mode
+'4' - Select control: AE compensation
+'5' - Select control: anti-banding/flicker mode
+'6' - Select control: effect mode
+'7' - Select control: brightness
+'8' - Select control: contrast
+'9' - Select control: saturation
+'0' - Select control: sharpness
+'[' - Select control: luma denoise
+For the 'Select control: ...' options, use these keys to modify the value:
+  '-' or '_' to decrease
+  '+' or '=' to increase
+
+'C' - capture a set of frames (PNG and unprocessed)
+'Q' - quit
+'''
+
+import cv2
+import numpy as np
+import numba as nb
+import depthai as dai
+
+streams = []
+# Enable one or both streams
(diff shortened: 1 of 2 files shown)


alex-luxonis:
image

chris-piekarski:
Another quick question for you @alex-luxonis. When running both ISP and RAW I am getting about 10fps for preview streams. If I disable RAW I can get 22-24 fps on ISP. But if I use cam.setFPS(60) it doesn't improve. Wondering if I am hitting USB bandwidth limit on the RPI3 or if the MonoCamera has other internal limits? If I drop to T_720_P from 800 I get another 2fps. Given the specs I was thinking I could get more than 24 fps out of it. Thoughts?

Luxonis-Brandon:
Yes, this is very likely a USB2 limitation/Pi CPU limitation.  I recall in Gen1 we had the capability to increase the size of data chunks that are transferred, which helps reduce CPU load and thereby increase throughput.  I'm not sure if it's in Gen2 yet though.

To tell if it's a USB bandwidth issue, you could run this example, which does encoding, and will reduce the throughput requirements by about an order of magnitude:
[https://docs.luxonis.com/projects/api/en/latest/samples/rgb_mono_encoding/#rgb-mono-encoding](https://docs.luxonis.com/projects/api/en/latest/samples/rgb_mono_encoding/#rgb-mono-encoding)

Could be missing something though, so definitely feel free to wait for Alex on this.  :-)

chris-piekarski:
Thanks @Luxonis-Brandon. I will run that example and then switch to RPI4 to see if FPS goes up.

chris-piekarski:
So must be bandwidth. I can get 60fps using single stream mono on Jetson/USB3. Pushing fps above 60 kicks out a system warning that fps is capped for 720p resolution.

Luxonis-Brandon:
Thanks.  And if you do lower resolution (640x400) you can do up to 118FPS on the grayscale.

klemen1999:
Closed due to inactivity

PR #138, opened 2021-05-30
Refactor deeplab depth

Erol444:
- Used color camera with rgb/depth alignment instead of the rectified right stream for NN inference
- New video:
depth-gif


Luxonis-Brandon:
Looks great!

VanDavv (approved):
LGTM

PR #139, opened 2021-05-31
Deeplab depth host-side sync

Erol444:
Added host-side NN sync, removed flickering
deeplabsync


Luxonis-Brandon:
Oh wow that looks great and way better!

VanDavv (approved):
LGTM, looks great!

PR #145, opened 2021-06-11
added gen2 cumulative object counting example

TannerGilbert:
Cumulative object counting for both RGB camera or external videos.

Based on:
* [Tensorflow 2 Object Counting repository](https://github.com/TannerGilbert/Tensorflow-2-Object-Counting)
* [OpenCV People Counter](https://www.pyimagesearch.com/2018/08/13/opencv-people-counter/) 
* [tensorflow_object_counting_api](https://github.com/ahmetozlu/tensorflow_object_counting_api)


gen2-cumulative-object-counting/centroid_tracker.py
```diff
@@ -0,0 +1,165 @@
...
+
+            # otherwise, if the number of input centroids is greater
+            # than the number of existing object centroids we need to
+            # register each new input centroid as a trackable object
+            else:
+                for col in unusedCols:
+                    self.register(inputCentroids[col])
+
+        # return the set of trackable objects
+        return self.objects
```
Erol444, line 165: From my understanding, this has the same functionality as ObjectTracker, just that it runs on the host?

Erol444 (changes requested):
Thank you for this demo, it will be useful to the community! The only thing I would change is to move the object tracking logic (if I understood it correctly) to the device using ObjectTracker node :)

TannerGilbert:
@Erol444 Didn't know there was a Tracker built into the depthai library. I just rewrote the code to use it instead of the CentroidTracker.

Luxonis-Brandon:
Sweet thanks @TannerGilbert !

Erol444:
Just tried it out, works great! Only thing I would change is to have the model set by default (to the `/models/mobilenet-ssd.blob`) and possibly adding the video from the readme (if it's not too big) to the `demo` folder. Thanks!

Erol444 (approved):

TannerGilbert:
@Erol444 Done

gen2-cumulative-object-counting/trackable_object.py
```diff
@@ -0,0 +1,11 @@
+# from https://www.pyimagesearch.com/2018/08/13/opencv-people-counter/
+class TrackableObject:
```
VanDavv, line 2: Could we move the declaration of this class to the `main.py`? It would be one file less and this class is compact anyways

TannerGilbert:
@VanDavv Moved it into main.py

VanDavv (approved):
LGTM, thanks!

PR #146, opened 2021-06-14
Add Gen2 WebRTC example

VanDavv:
# Gen2 WebRTC streaming example

This example demonstrates how to setup a WebRTC server to configure the device and stream the results and preview from it.

## Demo

[Gen2 WebRTC](https://youtu.be/8aeqGgO8LjY)

## Pre-requisites

1. Purchase a DepthAI model (see [shop.luxonis.com](https://shop.luxonis.com/))
2. Install requirements
   ```
   ```
3. [Enable insecure origins to be treated as secure (Chrome)](https://stackoverflow.com/a/58449078/5494277)

   > To ignore Chrome’s secure origin policy, follow these steps. Navigate to chrome://flags/#unsafely-treat-insecure-origin-as-secure in Chrome.  
   > Find and enable the Insecure origins treated as secure section (see below). Add any addresses you want to ignore the secure origin policy for. Remember to include the port number too (if required). 
   example
   > Save and restart Chrome. 
   > Remember this is for dev purposes only. The live working app will need to be hosted on https.

## Usage

```
python3 server.py
```

And open [`http://0.0.0.0:8080`](http://0.0.0.0:8080)

localhost preview

## Modify the script

DepthAI part of the code is stored in `transformators.py`, as `DepthAIVideoTransformTrack`.
You can add more capabilities there, like modify the pipeline or output.

If you'd like to send the nn results using datachannel, please use the following snippet inside `get_frame` method of the transformator

```python
if self.pc_id in self.application.pcs_datachannels:
    channel = self.application.pcs_datachannels[self.pc_id]
    channel.send(json.dumps({
        'type': 'NEW_RESULTS',
        'payload': [] # your results array here
    }))

```

If you'd like to add more config options to the script, first add a new input with a correct `name` attribute
to `client/index.html` inside `#options-form`. It will be automatically parsed and sent to the server.
There, you can access them in the transformer by either referncing `self.options.raw_options.get("<name_attribute>")`
or by adding a new property in `OptionsWrapper` class


Erol444 (approved):
LGTM, thanks :)

TannerGilbert:
Just tried it. Works well, even on Windows. @VanDavv thanks a lot for implementing it so quickly.

themarpe (changes requested):
The client/build/client.js seems like an build artifact.
Remove and add to gitignore

VanDavv:
@TannerGilbert is right, it's a generated file indeed but it's relatively small and keeping it compiled makes the initial setup much easier @themarpe.

themarpe (approved):
Sounds good then:)

PR #154, opened 2021-06-28
update license plates experiment to use latest depthai version and blobconverter

VanDavv:


Erol444 (approved):
thanks for updating this:)

PR #157, opened 2021-07-08
Extend support oak1

Erol444:


gen2-mjpeg-streaming/main.py
```diff
@@ -75,85 +75,95 @@ class ThreadedHTTPServer(ThreadingMixIn, HTTPServer):
...
+
+        xoutDepth = pipeline.createXLinkOut()
+        xoutDepth.setStreamName("depth")
+        mobilenet.passthroughDepth.link(xoutDepth.input)
+
+    return pipeline
 
 # Pipeline is defined, now we can connect to the device
-with dai.Device(pipeline) as device:
+with dai.Device(dai.OpenVINO.Version.VERSION_2021_2) as device:
```
VanDavv, line 152: Can we use blobconverter to fetch the blob with latest OpenVINO version?

VanDavv (approved):
Thanks! Despite a single comment regarding blob version - LGTM

gen2-mjpeg-streaming/main.py
Erol444: done, thanks for suggestion:)

PR #159, opened 2021-07-15
gen2-seq-num-sync: sync all 3 cams

alex-luxonis:
Using updated FW where RGB - Mono capture time sync is implemented. See related PRs:
https://github.com/luxonis/depthai-python/pull/310
https://github.com/luxonis/depthai-core/pull/174

```
usage: main.py [-h] [-f FPS] [-d] [-v] [-t]

optional arguments:
  -h, --help           show this help message and exit
  -f FPS, --fps FPS    Camera sensor FPS, applied to all cams
  -d, --draw           Draw on frames the sequence number and timestamp
  -v, --verbose        Verbose, -vv for more verbosity
  -t, --dev_timestamp  Get device timestamps, not synced to host. For debug

Press C to capture a set of frames.
```

Example output:
```
depthai-experiments/gen2-seq-num-sync$    python3 main.py -d -f 30 -vv

   Seq  Left_tstamp  RGB-Left  Right-Left  Dropped
   num    [seconds]  diff[ms]    diff[ms]  delta
     0     0.055592    -0.785       0.017
     1     0.088907    -0.771       0.011
     2     0.122222    -0.758       0.009
     3     0.155537    -0.745       0.010
     4     0.188852    -0.340       0.011
     5     0.222167    -0.326       0.011
     6     0.255482     0.063       0.010
     7     0.288796     0.078       0.010
     8     0.322111     0.257       0.010
     9     0.355426     0.270       0.009
    10     0.388741     0.246       0.010
    11     0.422056     0.260       0.011
    12     0.455371     0.146       0.009
    13     0.488686     0.160       0.009
    14     0.522000     0.046       0.010
    15     0.555315     0.061       0.010
    16     0.588644    -0.015       0.010
    17     0.621945     0.020       0.011
    18     0.655267    -0.011       0.008
    19     0.688575     0.019       0.010
    20     0.721890     0.018       0.009
    21     0.755205     0.033       0.009
    22     0.788520     0.037       0.012
    23     0.821834     0.050       0.010
    24     0.855149     0.048       0.009
    25     0.888463     0.063       0.010
    26     0.921778     0.055       0.012
...
```
EDIT: Adding more info:
Note that sometimes a jump of up to few milliseconds can be observed running as in the command above (or simplified with `-v` instead of `-vv`, to print a new line only when frames got dropped, or when `RGB-Left` timestamp diff is larger than 0.15ms). The jump seems to be related to the host-device time sync mechanism. By adding `-t` - that is using the raw device timestamps, the jump should not be observed.


alex-luxonis:
Capture saved when pressing `C` at runtime:
capture_0005

Luxonis-Brandon:
Sweet!

alex-luxonis:
(edited PR description to add more info on a potential time diff jump)

Erol444 (approved):
LGTM, thanks!

VanDavv (approved):
Looks great to me!

Richard-Elliott:
Hi, i'm just getting started with DepthAI, I found this bit of code very useful for the syncing of images. Can this be extended to sync bitstream videos so that each frame in the video is recorded at the same time, or that a single video contains all 3 camera feeds. 
Thank you

Erol444:
Hello @Richard-Elliott , you could combine all 3 frames into one single bigger frame with the help of OpenCV and save those frames into a video. Thoughts?

Richard-Elliott:
Hi @Erol444, the code I copied from here did already combine the synced frames, so saving those into a video seems logical, I will have a go at that. 
I was unsure of the difference between doing the video encoding on the host rather than the device. Is what you suggest possible on the device or is the video encoding going to be on the host? It would be nice to be able to create synced video streams from the device if that is possible.

Erol444:
Hello @Richard-Elliott, so encoding is quite computationally expensive, so shifting the heavy lifting to the device will reduce CPU usage on host. I'm not sure how you could combine 3 encoded streams to a single video, so I suggest to combine (un-encoded) streams on the host.

julianaddison:
Hi @alex-luxonis,

I'm trying to use the _PairingSystem_ class in gen2-seq-num-sync for my pipeline which has a rgb_preview, depth, and NN output. The rgb_preview is passed into the NN model which generates the NN output. My goal is to use the NN_output and get the depth values for it. This mean I need to get the same pairs for the NN_output and depth 

Assumed we could sync the NN output via the add_packet method but get the following error:

`AttributeError: 'depthai.NNData' object has no attribute 'getInstanceNum'`

Somewhat stuck in terms of how to proceed.  I'm not sure if I'm approaching this the right way. Instead I should be looking at but how to get the "frame/sequence id" for the rgb_preview which matches the NN output. Is there anyway to find out what is the rgb_preview packet/sequence number the NN_output belong to? Is there any example I could follow?

alex-luxonis:
Hi @julianaddison , 
`NNData` is missing some attributes that `ImgFrame` packets have, like camera timestamp / instance id, sequence number. We plan to improve this in the future, and forward these fields taken from the image frame inference ran on.

For now, it's possible to get synced pairs of the NNData result and the image frame it, by using the `passthrough` output of the NN node, see the usage of `--sync`/`args.sync` in this example:
https://github.com/luxonis/depthai-python/blob/3da450f/examples/rgb_mobilenet.py

Then it may be possible to use the `PairingSystem` class in this example with the `passthrough` ImgFrame and depth output, and the `passthrough` packet having associated to it the NNData packet (would need some code changes in the script to keep track of these).
Also note, if the NN is slower, for the `PairingSystem` implementation here to work well, you may need to reduce the FPS of all cameras to the rate NN can run - `.setFps(..)`.

PR #161, opened 2021-07-19
Update DepthAI API version

VanDavv:
Updates Gen2 experiments with latest Gen2 API version (`2.7.2.0`)

Current issues:
- https://github.com/luxonis/depthai-experiments/issues/162 -`gen2-deeplabv3_depth` experiment is not working with latest API version, sync mechanism is not updating depth frames even though they arrive. Potentially solvable by using #159 for sync
- https://github.com/luxonis/depthai-experiments/issues/163 - `gen2-gaze-estimation` yields incorrect gaze vector
- https://github.com/luxonis/depthai-experiments/issues/164 - `gen2-people-tracker` yields inconsistent results due to video FPS being too high for the nn to produce some valid data (but doesn't run at all with the version specified in the requirements so updated it too)
- `gen2-triangulation` API version cannot be updated until this PR is merged - https://github.com/luxonis/depthai-python/pull/316


Luxonis-Brandon:
Thanks.  We should go ahead and push those that do work in a wave so that folks can use these and it's less likely for a customer to have the bad UX of one not working, when really the only thing preventing it is an old requirementx.txt.  

CC: @Erol444

Erol444:
I agree with @Luxonis-Brandon, we can go ahead and merge the ones that are converted already :+1:

VanDavv:
Sure @Luxonis-Brandon and @Erol444, I created separate issues for each of the mentioned experiments and will focus on them separately

Erol444 (approved):
Thanks:) And it's great to remove all these blobs, we might want to delete the history afterwards (to reduce the size of the repo). Thoughts? CC @themarpe

VanDavv:
I think it would be good to do so rather sooner than later. Maybe once we migrate all examples to Gen2 would be a good time to do so?

Luxonis-Brandon:
Thanks both.  And yes don't let removing the blobs block getting these examples out/merged/working for POE.  Do as separate effort after this is out there and we are not causing pain for our customers.

Issue #166, opened 2021-07-22
gen2-ocr code not working openCV version 4.2.0 over


Thrada:
[gen2-ocr](https://github.com/luxonis/depthai-experiments/tree/master/gen2-ocr) doesn't work in opencv version over 4.2.0 with following issue

![image](https://user-images.githubusercontent.com/53900404/126597319-af99cf75-afb6-4c47-ba30-09d13500abad.png)

It caused because boxPoint doesn't recognize numpy datatype. 

I solved this problem just change numpy datype to normal datatype in [east.py](https://github.com/luxonis/depthai-experiments/blob/f06312596b47427f9699a300ae8bc143dbbac000/gen2-ocr/east.py#L13)

PR #167, opened 2021-07-22
Issue/166

Thrada:
solved problem issue #166


Erol444 (approved):
Just tried it out and it works (it worked for me before as well though I was using opencv version `4.5.1`. Thanks for the PR @Thrada!

PR #171, opened 2021-08-02
Script frame sequence number matching

Erol444:
2nd stage NN should now get the same frame as the 1st stage NN by using frames sequence number in the scripting node.

Working confirmed [on Discord](https://discord.com/channels/790680891252932659/795743863519576105/871795031781040148).

### DIFF
--- a/gen2-age-gender/main.py
+++ b/gen2-age-gender/main.py
@@ -86,6 +86,8 @@ l = [] # List of images
 # For this experiment this function is redundant, since everything
 # runs in blocking mode, so no frames will get lost
 def remove_prev_frame(seq):
+    if len(l) == 0:
+        return
     for rm, frame in enumerate(l):
         if frame.getSequenceNum() == seq:
             # node.warn(f"List len {len(l)} Frame with same seq num: {rm},seq {seq}")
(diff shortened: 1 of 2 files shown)


gen2-age-gender/main.py
```diff
@@ -74,36 +74,66 @@ def create_pipeline():
...
+l = [] # List of images
+# So the correct frame will be the first in the list
+# For this experiment this function is redundant, since everything
+# runs in blocking mode, so no frames will get lost
+def remove_prev_frame(seq):
+    for rm, frame in enumerate(l):
+        if frame.getSequenceNum() == seq:
+            # node.warn(f"List len {len(l)} Frame with same seq num: {rm},seq {seq}")
+            break
+    for i in range(rm):
```
VanDavv, line 93: on the first iteration, if `l` is empty, the `rm` variable would not be initialized and throw error

VanDavv (changes requested):
Nice,

gen2-age-gender/main.py
Erol444: Thanks and thanks for noticing this! I'm not sure how inferencing could have arrived to the script node before the frame, but better to ba safe, especially when other users might edit this script.

VanDavv (approved):
Just gave it a try using 2.9.0.0 API and works great, thanks!

PR #172, opened 2021-08-03
gen2-OCR experiment refactored

Erol444:
OCR cropping after detection is done on high quality (1024x1024) frame instead of passthrough frame (256x256)

wanted to add blobconverter but couldn't find `text-detection.blob` in the openvino model zoo. 
Edit: text-detection blob from [here](https://github.com/PINTO0309/PINTO_model_zoo/tree/main/051_East_Text_Detection).
PR in depthai model zoo [here](https://github.com/luxonis/depthai-model-zoo/pull/3).


Luxonis-Brandon:
Looks a lot better to me; thanks, @Erol444 .

One note:  
image

This had actually properly worked to read `altoids` but the _flashiness_ of the display causes the text to not show up in the screenshot.

Is it possible to make the text persistent and only disappear when updated?  As now for some reason it is made blank for some time period (maybe 50% duty cycle?) prior to new text being written.

Luxonis-Brandon:
image
Same here.  It correctly shows `marmot`, but since it flashes, it's very hard to catch it in a screenshot.

VanDavv:
@Luxonis-Brandon I gave it a try on latest DepthAI version and it seems to work correctly (although a bit slow)


https://user-images.githubusercontent.com/5244214/134329975-d91da66b-3e17-485a-9284-129c013285e4.mov

VanDavv (approved):

PR #187, opened 2021-08-24
Add checks and adjustments to run Gen2 experiments on OAK-1

VanDavv:


Erol444:
@VanDavv  this PR broke some experiments (eg. triangulation, deeplab) due to `Device booted with different OpenVINO version that pipeline requires`. Solution: specify the openvino version when creating device (otherwise it will take the latest), eg. `with dai.Device(pipeline.getOpenVINOVersion()) as device:`. Could you check this?

PR #193, opened 2021-09-02
Refactored stereo inference

Erol444:
refactored stereo inference demo so it uses disparity, then calculates depth/spatial coordiantes. Depth/Spatial coordinates seem to be quite accurate.

image


VanDavv (approved):
Just tested it on Mac - works great! One suggestion would be to add a cli flag `--visualize` instead of hardcoded bool value

PR #198, opened 2021-09-07
Added gen2-play-encoded-stream

MOj0:
Note: it has not been tested on Linux.


gen2-play-encoded-stream/main.py
```diff
@@ -0,0 +1,44 @@
...
+
+width, height = 720, 500
+cmd_out = ["ffplay", "-i", "-", "-x", str(width), "-y", str(height)]
+if osName == "nt":  # Running on Windows
+    cmd_out = ["cmd", "/c"] + cmd_out
+
+# Connect to device and start pipeline
+with dai.Device(pipeline) as device:
+    # Output queue will be used to get the encoded data from the output defined above
+    q = device.getOutputQueue(name="h265", maxSize=30, blocking=True)
```
themarpe, line 34: Set to 8 frames and non blocking
This will allow for slow ffplay consumption to drop frames instead of block and buildup

gen2-play-encoded-stream/main.py
```diff
@@ -0,0 +1,44 @@
...
+# Properties
+camRgb.setBoardSocket(dai.CameraBoardSocket.RGB)
+camRgb.setResolution(dai.ColorCameraProperties.SensorResolution.THE_4_K)
+videoEnc.setDefaultProfilePreset(3840, 2160, 30, dai.VideoEncoderProperties.Profile.H265_MAIN)
+
+# Linking
+camRgb.video.link(videoEnc.input)
+videoEnc.bitstream.link(xout.input)
+
+width, height = 720, 500
```
themarpe, line 26: To what width/height is this related?
(Check camRgb.getVideoSize / getVideoWidth/Height as well as for setDefaultProfilePreset)

themarpe (changes requested):

gen2-play-encoded-stream/main.py
MOj0: This is used only for the ffplay command (-x and -y arguments).

gen2-play-encoded-stream/main.py
MOj0: By setting `maxSize=8` and `blocking=False` I get a weird bug:

image

Console output:
`[hevc @ 0000019747f6f3c0] Stream #0: not enough frames to estimate rate; consider increasing probesize`
`[hevc @ 000001974e9b7a00] Could not find ref with POC 22`

themarpe: @MOj0 frames are dropping in this case - you can increase keyframe frequency to reduce artifacts when frames drop (VideoEncoder), and/or set to h264 encoding as is usually easier to decode (more HW support)

Might also make sense to increase maxSize to a higher value - 8 might be too little, depending on how ffplay consumes data.

gen2-play-encoded-stream/main.py
themarpe: This being the final displayed window size then?

MOj0: Yes, this is used just to lower the ffplay resolution, otherwise it opens a 3840x2160 window.

Erol444 (approved):
lgtm :+1:

themarpe (approved):

PR #202, opened 2021-09-08
Replay/Record feature

Erol444:
IMO it's good enough for now, we will be adding new features later.

TODO:
- support recording depth into ROS bags, so you can view depth stream with Realsense Viewer
- Allow greater configurability
- support recording IMU (also ROS bag)
- Record directly to eg. `.avi` instead of the `.mjpeg` so the video can be viewed with any video player
- Add these 2 libraries to the `depthai_sdk`


themarpe (approved):
Not a general solution in this case, but we lack some things to make it happen (Decoupled camera & ISP & post processing step)

Also, currently you save `video` which type is NV12 but replay as BGR888p. Just a heads up for a potential general solution

VanDavv (approved):
Thanks @Erol444, would be nice to integrate this into the demo script and SDK at some point

PR #204, opened 2021-09-10
Add gen2-face-detection example

tersekmatija:
Added gen2-face-detection example, NN has input shape 160x120. Camera preview size is set to 640 x 480 and is resized with ImageManip to input shape. Result is shown on the camera preview, so accuracy might be a bit worse, but will add NN with bigger input shapes in the future.


Luxonis-Brandon:
The GIF makes me think "How you'll feel when you back the OAK-D-Lite KickStarter"

Erol444 (approved):

PR #205, opened 2021-09-11
Face recognition refactor

Erol444:
Everything now runs on-device except face vector comparison (in theory could also do it on device).
**TODO:**
- add readme
- add gif
- specify requirements.txt (latest dev branch of lib)

**Note:**
- there is no frame-NN sync so bounding boxes aren't in-sync (due to 3 models being used)
- recognizing 2 faces at the same time might not work 100%

**Demo gif**
ezgif com-gif-maker(1)


VanDavv (approved):
The demo is very fun to play! 
I'll see if I have some time in the following days to fix the issue with detecting two faces, but for a refactor I think it's ok to leave it as is.
One suggestion would be not to draw detected names on the frame when the `--name` parameter is specified.

Other than that, LGTM

PR #209, opened 2021-09-14
Added gen2-depth-mbnv2

tersekmatija:
Added gen2-depth-mbnv2 to the experiments.


Erol444 (approved):
Awesome, thanks! I would just like to mention that we should be saving these .blob files (better yet xml/bin) somewhere in the cloud and use blobconverter to download & convert them (since these to add 60MB to the repo). CC @VanDavv

Erol444 (changes requested):
Use blobconverter and depthai-model-zoo to download the model

Erol444 (approved):

Erol444:
example

PR #211, opened 2021-09-19
gen2-human-machine-safety: improvements

ES-Alexander:
- improved clarity by reducing copied code and improving variable names
- vectorised `calc_spatials` depth averaging
- added easier support for alternative averaging methods (e.g. `np.median`)

### DIFF
--- a/gen2-human-machine-safety/main.py
+++ b/gen2-human-machine-safety/main.py
@@ -25,11 +25,16 @@ def crop_to_rect(frame):
     # print(height, width, delta)
     return frame[0:height, delta:width-delta]
 
+def annotate(img, text, pos, color, fontFace=cv2.FONT_HERSHEY_TRIPLEX, fontScale=0.5, **kwargs):
+    cv2.putText(img, text, pos, fontFace, fontScale, color, **kwargs)
+
+def bold(x, level=2):
+    return range(x, x+level)
+
+
 class HumanMachineSafety:
     def __init__(self):
         print("Loading pipeline...")
-        self.fontScale = 1
-        self.lineType = 0
         self.palmDetection = PalmDetection()
 
         self.distance = 1000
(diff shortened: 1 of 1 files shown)


gen2-human-machine-safety/main.py
```diff
@@ -39,31 +44,26 @@ def __init__(self):
...
-        averageDepth = sum / cnt if 0 < cnt else 0
+        deltaX = int((xmax - xmin) * 0.33)
+        deltaY = int((ymax - ymin) * 0.33)
+        xmin += deltaX
+        ymin += deltaY
+        xmax -= deltaX
+        ymax -= deltaY
+
+        # Calculate the average depth in the ROI.
+        depthROI = depth[ymin:ymax, xmin:xmax]
```
VanDavv, line 58: Sometimes the detection bbox is flipped (e.g. `bbox` variable can be `[720, 720, 0, 0]`), so cutting the frame before check throws an error at random during runtime

VanDavv (changes requested):
@ES-Alexander Thanks for your PR! I found some issues, but your implementation on depth ROI calculation is definitely better. I went ahead quickly and self-addressed the comments below in this PR - https://github.com/luxonis/depthai-experiments/pull/212 - where you can find both your commit and my improvements

Luxonis-Brandon:
Thanks @ES-Alexander and @VanDavv !

ES-Alexander:
> I found some issues

Fair enough - I probably should have been more rigorous with, you know, actually testing it :-P
The depth averaging was linked to from somewhere (can't remember if it was discord or in the OAK-D-Lite kickstarter FAQs or something). I swung by to take a look and saw that it could be done better, and once I'd fixed that I saw a couple of other repetitive things that I thought I may as well fix while I was in there :-)

> I went ahead quickly and self-addressed the comments below in this PR - #212 - where you can find both your commit and my improvements

Thanks!

PR #212, opened 2021-09-20
Improve human machine safety demo

VanDavv:
This PR is an updated version of #211, containing both the changes from the mentioned PR and a couple of bug/style fixes

### DIFF
--- a/gen2-human-machine-safety/main.py
+++ b/gen2-human-machine-safety/main.py
@@ -25,11 +25,11 @@ def crop_to_rect(frame):
     # print(height, width, delta)
     return frame[0:height, delta:width-delta]
 
-def annotate(img, text, pos, color, fontFace=cv2.FONT_HERSHEY_TRIPLEX, fontScale=0.5, **kwargs):
-    cv2.putText(img, text, pos, fontFace, fontScale, color, **kwargs)
 
-def bold(x, level=2):
-    return range(x, x+level)
+def annotate_fun(img, color, fontFace=cv2.FONT_HERSHEY_TRIPLEX, fontScale=0.5, **kwargs):
+    def fun(text, pos):
+        cv2.putText(img, text, pos, fontFace, fontScale, color, **kwargs)
+    return fun
 
 
 class HumanMachineSafety:
@@ -53,6 +53,14 @@ class HumanMachineSafety:
         ymin += deltaY
         xmax -= deltaX
         ymax -= deltaY
+        if xmin > xmax:  # bbox flipped
+            tmp = xmin
+            xmin = xmax
+            xmax = tmp
+        if ymin > ymax:  # bbox flipped
+            tmp = ymin
+            ymin = ymax
+            ymax = tmp
 
         # Calculate the average depth in the ROI.
         depthROI = depth[ymin:ymax, xmin:xmax]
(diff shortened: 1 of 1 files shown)


Erol444:
Could we actually use this helper for text visualization: https://github.com/luxonis/depthai-experiments/blob/master/gen2-triangulation/main.py#L14-L22

gen2-human-machine-safety/main.py
```diff
@@ -39,31 +44,34 @@ def __init__(self):
...
+        deltaX = int((xmax - xmin) * 0.33)
+        deltaY = int((ymax - ymin) * 0.33)
+        xmin += deltaX
+        ymin += deltaY
+        xmax -= deltaX
+        ymax -= deltaY
+        if xmin > xmax:  # bbox flipped
+            tmp = xmin
+            xmin = xmax
+            xmax = tmp
```
ES-Alexander, line 59:
```suggestion
            xmin, xmax = xmax, xmin
```

gen2-human-machine-safety/main.py
```diff
@@ -39,31 +44,34 @@ def __init__(self):
...
+        if ymin > ymax:  # bbox flipped
+            tmp = ymin
+            ymin = ymax
+            ymax = tmp
```
ES-Alexander, line 63:
```suggestion
            ymin, ymax = ymax, ymin
```

ES-Alexander (approved):
Temp variables unnecessary in Python because of iterable unpacking.

Other than that all seems good to me - thanks for fixing this up :-)

ES-Alexander:
@Erol444 
> Could we actually use this helper for text visualization: https://github.com/luxonis/depthai-experiments/blob/master/gen2-triangulation/main.py#L14-L22

Just took a look at this and it seems unnecessarily awkward to change parameters in it. Maybe better to use a dataclass?

```python3
from dataclasses import dataclass
from typing import Tuple

@dataclass
class TextHelper:
    bg_color: Tuple[int, int, int] = (0, 0, 0)
    color: Tuple[int, int, int] = (255, 255, 255)
    text_type: int = cv2.FONT_HERSHEY_SIMPLEX
    text_scale: float = 0.5
    line_type: int = cv2.LINE_AA

    def putText(self, frame, text, coords, thickness=1):
        cv2.putText(frame, text, coords, self.text_type, self.text_scale, self.bg_color, 3 + thickness, self.line_type)
        cv2.putText(frame, text, coords, self.text_type, self.text_scale, self.color, thickness, self.line_type)
```

That way it can be used with the defaults, but is also easy to set up with different options, e.g.
```python3
annotator = TextHelper(color=(255, 0, 0), text_scale=3)
annotator.putText(frame, "something normal", (15, 20))
annotator.putText(frame, "something bold", (15, 40), 2)
```

gen2-human-machine-safety/main.py
VanDavv: ha, it was one of the first nuances about Python I've learned, but obviously, I forgot about it over time, thanks for pointing it up!

VanDavv:
I think that all of the approaches for "text helpers" are just different permutations and don't change much in the code itself, we still just want to print a text, omitting repetitive arguments.
@Erol444 I think we should add the TextHelper you mentioned to SDK and then use it across experiments

ES-Alexander (approved):

Erol444:
Maybe the way @ES-Alexander mentioned (as I was not aware of `@dataclass`), so it's more configurable:)

Erol444 (approved):

PR #213, opened 2021-09-20
Added mesh I/O functionality

MOj0:


saching13:
@MOj0  I think we can squash this since it is copy of what we moved this to `depthai-python`

PR #219, opened 2021-09-23
Added gen2-mobile-object-localizer

tersekmatija:
Added gen2-mobile-object-localizer


Erol444:
ezgif com-gif-maker(2)

PR #221, opened 2021-09-25
Refactored gen2-head-posture-detection

MOj0:


Erol444:
Refactored here: https://github.com/luxonis/depthai-experiments/pull/362

PR #222, opened 2021-09-25
specify pipeline in the Device constructor

harajune:
I encountered the error below.
> RuntimeError: Device booted with different OpenVINO version that pipeline requires

The error seems to occur if the previous device firmware is different from the model version.


Erol444 (changes requested):
Actually, a better solution here would be to specify openVino version - since we are passing the pipeline after we check if we even can run the pipeline (if device has mono cameras).

harajune:
Thank you for your review!
I modified the code as you commented.

Erol444 (approved):
Thank you!

PR #225, opened 2021-09-30
Changed first stage NN in triangulation demo to MobilenetDetectionNet…

Erol444:
…work, so there's no need for parsing in Script node


VanDavv (approved):
LGTM

PR #226, opened 2021-09-30
Record/replay fix so it works on Windows

Erol444:
Fixed:
- Problem with local paths (now uses `pathlib`)
- Issue with pickle-ing dai.Device - moved function to save frames (in new thread) outside the class
- Calling same function multiple times for some reason

### DIFF
--- a/gen2-record-replay/libraries/depthai_record.py
+++ b/gen2-record-replay/libraries/depthai_record.py
@@ -2,7 +2,41 @@
 from pathlib import Path
 from multiprocessing import Process, Queue
 from cv2 import VideoWriter, VideoWriter_fourcc
+import os
 import depthai as dai
+
+def store_frames(path, frame_q):
+    files = {}
+
+    def create_video_file(name):
+        file_path = str(path / f"{name}.mjpeg")
+        # fourcc = VideoWriter_fourcc(*'MJPG')
+        # width = self.nodes[name].getResolutionWidth()
+        # height = self.nodes[name].getResolutionHeight()
+        # writer = VideoWriter(path, fourcc, self.fps, (width, height))
+        # writer.release()
+        # time.sleep(0.001)
+        files[name] = open(file_path, 'wb')
+
+    while True:
+        try:
+            frames = frame_q.get()
+            if frames is None:
+                break
+            for name in frames:
+
+                if name not in files: # File wasn't created yet
+                    create_video_file(name)
+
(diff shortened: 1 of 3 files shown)


gen2-record-replay/libraries/depthai_record.py
```diff
@@ -2,8 +2,42 @@
...
+
+    def create_video_file(name):
+        file_path = str(path / f"{name}.mjpeg")
+        # fourcc = VideoWriter_fourcc(*'MJPG')
+        # width = self.nodes[name].getResolutionWidth()
+        # height = self.nodes[name].getResolutionHeight()
+        # writer = VideoWriter(path, fourcc, self.fps, (width, height))
+        # writer.release()
+        # time.sleep(0.001)
+        files[name] = open(file_path, 'wb')
```
VanDavv, line 19: Could you use [ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) instead? It's better for handling files

```
with ExitStack() as stack:
    for mgr in ctx_managers:
        stack.enter_context(mgr)
    # ...
```

gen2-record-replay/record.py
```diff
@@ -18,8 +20,8 @@
 parser.add_argument('-f', '--fps', type=float, default=30,
                     help='Camera sensor FPS, applied to all cams')
 # TODO: make camera resolutions configrable
-
 args = parser.parse_args()
+save_path = Path(__file__).parent.absolute() / args.path
```
VanDavv, line 24: This will make sure we're saving relatively to user's current directory, not the script itself

```suggestion
save_path = Path.cwd() / args.path
```

gen2-record-replay/record.py
```diff
@@ -51,53 +53,60 @@ def check_sync(queues, timestamp):
...
+                break
 
-    for recording in recordings:
-        recording.frame_q.put(None)
-        time.sleep(0.01) # Wait 10ms for process to close all video files
-        recording.process.join() # Terminate the process
+        for recording in recordings:
+            recording.frame_q.put(None)
+            time.sleep(0.01) # Wait 10ms for process to close all video files
+            recording.process.join() # Terminate the process
```
VanDavv, line 108: `join()` actually doesn't terminate the process, just hangs until it finishes - so the sleep before seems unnecessary

VanDavv (changes requested):

PR #229, opened 2021-10-07
Changed requirements to depthai 2.11.1.0

MOj0:
NOTE: gen2-social-distancing throws the following error, with both depthai 2.7.2.0 and 2.10.0.0 (tested on Windows): `RuntimeError: Device booted with different OpenVINO version that pipeline requires`

### DIFF
--- a/gen2-social-distancing/requirements.txt
+++ b/gen2-social-distancing/requirements.txt
@@ -2,4 +2,4 @@ imutils==0.5.3
 concurrent-log-handler==0.9.16
 opencv-python==4.5.1.48
 depthai==2.11.1.0
-blobconverter==1.0.0
+blobconverter==1.2.2
--- a/gen2-age-gender/requirements.txt
+++ b/gen2-age-gender/requirements.txt
@@ -1,5 +1,4 @@
 opencv-python==4.5.1.48
 imutils==0.5.4
-blobconverter==1.0.0
---extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
+blobconverter==1.2.2
 depthai==2.11.1.0
--- a/gen2-class-saver-jpeg/requirements.txt
+++ b/gen2-class-saver-jpeg/requirements.txt
@@ -1,3 +1,3 @@
 opencv-python
 depthai==2.11.1.0
-blobconverter==1.0.0
+blobconverter==1.2.2
--- a/gen2-cumulative-object-counting/requirements.txt
+++ b/gen2-cumulative-object-counting/requirements.txt
@@ -2,4 +2,4 @@ numpy==1.20.3
 opencv_python==4.5.1.48
 depthai==2.11.1.0
 scipy==1.4.1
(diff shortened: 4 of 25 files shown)


gen2-gaze-estimation/main.py
```diff
@@ -1,4 +1,5 @@
 import argparse
+from os import pipe
```
Erol444, line 2: any reason for this?

gen2-depth-driven-focus/main.py
```diff
@@ -59,11 +59,11 @@ def create_pipeline():
     controlIn.out.link(cam.inputControl)
 
     left = pipeline.createMonoCamera()
-    left.setResolution(dai.MonoCameraProperties.SensorResolution.THE_480_P)
+    left.setResolution(dai.MonoCameraProperties.SensorResolution.THE_400_P)
```
Erol444, line 62: this is actually tricky one, and I'm not sure how we will resolve it. Thing is that OAK-D-Lite has sensor resolution of 480P, while all other models have 800,720 and 400P.

gen2-social-distancing/requirements.txt
```diff
@@ -1,5 +1,5 @@
 imutils==0.5.3
 concurrent-log-handler==0.9.16
 opencv-python==4.5.1.48
-depthai==2.7.2.0
+depthai==2.10.0.0
```
Erol444, line 4: If gen2-social-distancing throws error, why would you update it?

Erol444 (changes requested):

gen2-depth-driven-focus/main.py
VanDavv: I believe there was a plan on adding some common options for sensor resolution, CC @alex-luxonis @themarpe

gen2-social-distancing/requirements.txt
MOj0: Because it throws an error in both versions, so I suspect the depthai version isn't the problem.

gen2-gaze-estimation/main.py
MOj0: I'm not sure, since I'm not the one who put that there. Removed it, since it's unnecessary.

gen2-social-distancing/requirements.txt
Erol444: so.. will you try to find the root cause of the problem?

gen2-gaze-estimation/main.py
MOj0: Hmm... I might have hit some key combination to import that, or VSCode did it automatically, not sure. Either way it's irrelevant.

gen2-depth-driven-focus/main.py
themarpe: IMO lets do OAK-D for now and once we pin down Lite interoperability in this usecase we can change

themarpe (changes requested):
Wait for 2.11.0 release

PR #233, opened 2021-10-12
Update blobconverter (1)

alex-luxonis:
Update blobconverter in `gen2-people-counter` and `gen2-people-tracker` to fix model downloading failure.

@Erol444 Could you take a look at `gen2-people-tracker`, I'm getting inconsistent results from run to run with the default `python3 main.py` (on the static video input), Y counter ends up either 0, 1, -1...


alex-luxonis:
Tested locally, merging for now.

PR #242, opened 2021-11-05
Added face emotions estimation

Erol444:
image


VanDavv (approved):
Just tried, LGTM. Would be nice to extend the SDK capabilities to be able to use it here (2 stage nn and custom script node)

PR #244, opened 2021-11-11
Display detections differet possibilities

Erol444:
1. Passthroug
passthrough

2. Squeeze (whole FOV)
squeeze

3. Crop video (higher resolution)
crop_video

4. Edit bounding boxes, display video
edit_bb


tersekmatija (approved):
2, 3, and 4 didn't work for me due to 0 division in FPSHandler. I added + 1 to timestamp on initialization, which does the trick.

I approved the PR, but in case we do not need `==` in requirements, please update them. 

Otherwise, looks good! 💯 Thanks!

PR #253, opened 2021-12-08
Added full fov NN inferencing demo

Erol444:
This demo shows how you can run NN inferencing on full FOV frames
image


VanDavv (approved):
LGTM

PR #256, opened 2021-12-08
Removed all blobconverter Path to string conversion, as it's not needed

Erol444:
Removed all blobconverter Path to string conversion, as it's not needed anymore


VanDavv (approved):
I'm so happy with this change, this string conversion was always a downside of using `pathlib` in terms of clearness. LGTM

Erol444:
I think it's been quite some time since setBlobPath automatically converts to string @VanDavv

PR #258, opened 2021-12-13
Deeplab sync fix

Erol444:
Sync fix, as reported by @gespona


gespona:
Hi !  LGTM (Implementation to sync pretty similar to mine on point cloud)
When enabling subpixel I notice similar behaviour (fps drop) 
Captura de pantalla (17)

Erol444:
Will create new PR to optionally choose sequence number syncing instead of timestamp syncing.

PR #261, opened 2021-12-15
Updated & tested all experiments

Erol444:
to depthai 2.13.3 so all experiments are supported by the OAK-D-Lite


tersekmatija (approved):
Thanks! We should really look into adding one general requirements that would work for most projects, and then additional requirements.txt for edge cases in projects directories.

Erol444:
I agree, let's sync offline on that.

PR #262, opened 2021-12-15
Added object detection running on multiple devices demo

Erol444:
image


VanDavv (approved):
Just tested, works great, thanks @Erol444 !

PR #263, opened 2021-12-16
Added spatial calculator example on host side

Erol444:
ezgif com-gif-maker(14)


VanDavv (commented):
I think the incorrect depth is shown incorrectly - `nanm`

Screenshot 2021-12-17 at 13 09 21

PR #268, opened 2021-12-26
Update to support multiple faces per image

madgrizzle:
I updated this demo to support multiple faces per camera frame.  I had to update the script so that the headpose section didn't pop() any images in the buffer because a second face will need that image as well.  The buffer is simply maintained at 18 frames.  The key though was to set both the headpose_manip and face_rec_manip nodes to wait for a config input.  Without this, the ImageManip nodes appeared to generate an output from something (looked like maybe the camera preview feed) and therefore everything became out of sync.  More notes where added to the code to help the reader understand how it works and what to watch out for.  It seems to run best at 5 FPS (at 10 FPS it seems to lose sync.. see the notes)

### DIFF
--- a/gen2-face-recognition/main.py
+++ b/gen2-face-recognition/main.py
@@ -1,4 +1,48 @@
 # coding=utf-8
+
+'''
+This demo consists of a pipeline that performs face recognition on the camera input with little
+host side processing (almost all on device using script node)
+
+The pipeline, in general, does the following (not all nodes are described):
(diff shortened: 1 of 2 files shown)


madgrizzle:
Also, I'm running this on 2.13.3-image_manip_refactor and don't have issues.  Not sure why it needed to get reverted to 2.10.

Erol444:
Should already be implemented (for some time now). Closing due to inactivity.

PR #271, opened 2021-12-31
Fix crash

kitfit-mikkel:
Was crashing with error:

```bash
TypeError: unsupported operand type(s) for /: 'int' and 'NoneType'
```

On Windows 10


Luxonis-Brandon:
Thanks!  CC: @saching13

PR #274, opened 2022-01-12
Simplify and update gen2 rgb depth align experiment

jjd9:
The gen1 rgbd example looked promising, but it seems like the push to migrate this to gen2 lost steam or the original dev got busy. I had to write this for my own use and wanted to share my work since I benefited so much from the existing depthai examples.

This boils down the experiment into two files:
1. visualize depth or rgbd image as a 2d image and 3d pointcloud (supports gen2's aligning on device option as well as aligning on the host. Not sure why I get those weird lines aligning on host though :thinking: )
2. extension of the projector_3d.py utility for doing 1 and 2 that supports downsampling the depth image / pointcloud since all those depth to 3d deprojections can be heavy on old laptops like mine :sweat_smile: .

Also, depending on the lighting, the oak-d-lite struggles to estimate depth on my plain white wall (regardless of how much post-processing I throw at it), but I think that is just a passive stereo issue. However, I really like that it knows that the depth is bad and thus has a low confidence score allowing me to filter it out :smiley: 

Result for 1:
pointcloud_only

Result for 2 (align on device):
align_on_device

pcl_align_on_device

Result for 2 (align on host):
align_on_host

pcl_align_on_host


Luxonis-Brandon:
Thanks.  So actually I think we do have a better version on Gen2 already.  But I think this was probably a great way to learn/practice all the math, so worth it anyway.

CC: @saching13 for any comments here.

saching13:
Hi,
Looks like you improved on the existing [PR](https://github.com/luxonis/depthai-experiments/pull/203).  
Thanks alot for the additional improvements. Will review it soon.

Luxonis-Brandon:
Ah cool thanks!

gen2-rgbd-projection/main.py
```diff
@@ -0,0 +1,330 @@
...
+############################################################################
+
+if (not lrcheck and not align_on_host) and align_depth:
+    print("Warning: Cannot align on device without lr check!")
+    align_on_host = True
+
+# Select camera resolution (oak-d-lite only supports THE_480_P  and THE_400_P depth)
+# res = {"height": 400, "width": 640,
+#        "THE_P": dai.MonoCameraProperties.SensorResolution.THE_400_P}
+res = {"height": 480, "width": 640,
```
saching13, line 51: any specific reason why we are assigning only OAK-D-Lite Res ?

gen2-rgbd-projection/main.py
```diff
@@ -0,0 +1,330 @@
...
+
+    # Connect to device and start pipeline
+    print("Opening device")
+    with dai.Device(pipeline) as device:
+        # get the camera calibration info
+        calibData = device.readCalibration()
+        left_intrinsic = np.array(calibData.getCameraIntrinsics(
+            dai.CameraBoardSocket.LEFT, res["width"], res["height"]))
+        right_intrinsic = np.array(calibData.getCameraIntrinsics(
+            dai.CameraBoardSocket.RIGHT, res["width"], res["height"]))
```
saching13, line 210: these would be wrong if you use the code with OAK-D. because OAK-D defaults to `400p` res when 480p is passed

gen2-rgbd-projection/main.py
```diff
@@ -0,0 +1,330 @@
...
+        _, rgb_default_width, rgb_default_height = calibData.getDefaultIntrinsics(
+            dai.CameraBoardSocket.RGB)
+        rgb_orig_intrinsic = np.array(
+            calibData.getCameraIntrinsics(dai.CameraBoardSocket.RGB))
+        # update camera intrinsics based on new image size
+        rgb_intrinsic = rgb_orig_intrinsic.copy()
+        width_scaling = res["width"] / rgb_default_width
+        height_scaling = res["height"] / rgb_default_height
+        rgb_intrinsic[0, :] *= width_scaling
+        rgb_intrinsic[1, :] *= height_scaling
```
saching13, line 222: focal length will also be different when you scale it. use `calibData.getCameraIntrinsics(dai.CameraBoardSocket.RGBm resize_width, resized_height)`

gen2-rgbd-projection/main.py
```diff
@@ -0,0 +1,330 @@
...
+                continue
+            elif pcl_frames[0] is None:
+                print("Waiting on depth image to visualize")
+                continue
+
+            if align_depth:
+                # Convert nonzero depth into red pixels in 3-channel image
+                depth_three_channel = np.zeros_like(pcl_frames[1])
+                depth_three_channel[:, :, 2] = (
+                    255 * pcl_frames[0].astype(float) / max_disparity).astype(np.uint8)
```
saching13, line 308: is this for visualization ?

gen2-rgbd-projection/main.py
```diff
@@ -0,0 +1,330 @@
...
+                    0.4 * pcl_frames[1].astype(float) / 255
+                blended_image = (255 * blended_image.astype(float) /
+                                 blended_image.max()).astype(np.uint8)
+                cv2.imshow("RGBD Alignment", blended_image)
+                if not no_pcl:
+                    pcl_converter.rgbd_to_projection(
+                        *pcl_frames, downsample=downsample_pcl)
+                    pcl_converter.visualize_pcd()
+            else:
+                quantized_depth_image = quantizeDisparityFrame(
```
saching13, line 322: `quantizeDisparityFrame` Change it to depth. We are not using disparity anywhere in this code.

saching13 (changes requested):

gen2-rgbd-projection/main.py
jjd9: That is my bad. I have an oak-d-lite, and I had hard-coded the desired resolution. The latest commit fixes this by still allowing the user to set the resolution, but now I get the height/width from the rgb and depth nodes to avoid hard-coding.

gen2-rgbd-projection/main.py
jjd9: The above mentioned commit fixes this. I also made it so that the code will exit and warn the user if it detects that the user-specified resolution does not match the resolution of images coming from the device (which would suggest the resolution setting is not compatible with the device).

gen2-rgbd-projection/main.py
jjd9: Yes, purely for visualization. here I am turning the depth image into a 3-channel image (where the depth values > 0 are red pixels and all others are white) so it can be overlaid onto the rgb image. This allows the user to visually verify the rgb-d alignment (since it can be cumbersome to do by looking at the pointcloud).

gen2-rgbd-projection/main.py
jjd9: Correct, typo on my part. That should be depth. Fixed :+1:

gen2-rgbd-projection/main.py
jjd9: This scales the first and second row of the rgb_intrinsic matrix (which would include the x/y focal lengths). I believe that getCameraIntrinsics does not behave properly if resize_width / original_rgb_width != resize_height / original_rgb_height (i.e. if the resize causes a change in image aspect ratio). That is why I am scaling the focal length and optical center manually here.

jjd9: :point_up: This theory is based on my testing (using getCameraIntrinsics as you suggested does not work properly) and from reading the source code, which does not seem to handle non-fixed aspect ratio resizing: https://github.com/luxonis/depthai-core/blob/8ab5bf954e3b65f6ec0933e372777aac10b81900/src/device/CalibrationHandler.cpp#L197 But please let me know if I am off base here.

gen2-rgbd-projection/main.py
jjd9: https://github.com/luxonis/depthai-experiments/blob/1d9fe1084ba613824630efdb165dbc0b3a00e69b/gen2-rgbd-projection/main.py#L196

gen2-rgbd-projection/main.py
jjd9: https://github.com/luxonis/depthai-experiments/blob/1d9fe1084ba613824630efdb165dbc0b3a00e69b/gen2-rgbd-projection/main.py#L129

gen2-rgbd-projection/main.py
jjd9: Also moving these utility parts of the code to functions in a separate file to de-clutter.

gen2-rgbd-projection/main.py
saching13: I think you are right. I just scales based on one of them. If possible feel free to create a PR there. Or else I will add it to my tasks. Sorry for the delay.

PR #281, opened 2022-01-21
Depth and disparity pipelines fixed

SergioVelmay:
I noticed that the queue called `depth` was never found.

This was the reason:
`"left" or "disparity" or "depth"` is always equal to `"left"`
`"right" or "disparity" or "depth"`  is always equal to `"right"`
`"disparity" or "depth"`  is always equal to `"disparity"`

The `depth` recording only worked if `left`, `right` and `disparity` were also selected.


Erol444 (approved):
Thank you for the bug fix! LGTM

PR #283, opened 2022-01-31
Add Qt GUI example

VanDavv:
# Gen2 QT GUI

This demo uses [PyQt5](https://pypi.org/project/PyQt5/) to create an interactive GUI displaying frames from DepthAI camera. 
Application allows switching between RGB and Depth previews and adjusting depth properties dynamically, like enabling Left-Right Check or changing median filtering.
Additionally, layout is built using [QML](https://doc.qt.io/qt-5/qtqml-index.html)

## Demo

depth gif

## Install project requirements

```
```

## Run this example

```
python3 main.py
```


Erol444 (approved):
LGTM, thanks!

PR #294, opened 2022-02-09
Update gen2-face-recognition to 2.14.1.0; mitigate script node crash by changing target CPU to CSS

szabi-luxonis:


Erol444 (approved):
Have been testing for 20min and seems to work flawlessly, thanks @szabi-luxonis !

PR #318, opened 2022-03-04
Full FoV NN - use latest depthai release

alex-luxonis:
Send properly scaled `preview` (812x608) to ImageManip.
Reduce FPS (30 -> 25) to fix ImageManip lag with ISP running at 12MP.


Erol444 (approved):
For some reason, I was under the impression the image_manip_refactor branch was needed for this functionality, thanks for fixing it!

PR #319, opened 2022-03-07
Created foxglove folder

ZigiMigi:
Created foxglove folder and 2 files. First streams rgb image to foxglove studio, second streams pointclouds to foxglove studio.


Luxonis-Brandon:
Sweet thanks @ZigiMigi !

ZigiMigi:
- when testing ```point cloud``` in foxglove you will see that the ```point cloud``` is flickering, that is not a problem on our side, but on foxglove's side (as they have a problem with their 3D panel)

Erol444 (changes requested):
1. Running `fosglove_pcl_stream.py` throws
 ```
Traceback (most recent call last):
  File "fosglove_pcl_stream.py", line 54, in <module>
    device = depthai.Device("", False)
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
```
I suspect you are using gen1 depthai? Migrate to gen2 (2.15.0)
2. Running `foxglovePC.py` throws 
```
  File "foxglovePC.py", line 112, in main
    "width": len(ball),
TypeError: object of type 'open3d.cuda.pybind.geometry.PointCloud' has no len()
```
3. Add `requirements.txt`
4. Move all logic into the same `main.py` file, and add CMD arguments so users can easily select what to stream (either pointcloud/RGB/left/right). By default stream rgb and pointcloud.
5. Add `readme.md` with tutorial how to get started.
6. Remove sphere.ply as it's not needed.

ZigiMigi:
pic4

Erol444:
Fixed argument setting, new PR here: https://github.com/luxonis/depthai-experiments/pull/370

PR #321, opened 2022-03-09
Added .mcap format to recording

ZigiMigi:
Added class Mcap to depthai_record and  added option to record files into .mcap format

### DIFF
--- a/gen2-record-replay/libraries/depthai_record.py
+++ b/gen2-record-replay/libraries/depthai_record.py
@@ -9,6 +9,7 @@ from multiprocessing import Queue
 from pathlib import Path
 from threading import Thread
 
+from depthai_mcap import DepthAiMcap
 import depthai as dai
 import numpy as np
 from mcap.mcap0.writer import Writer
@@ -43,7 +44,8 @@ class Record:
 
     def run(self):
         files = {}
-        mcaps = {}
+        if self.mcap:
+            mcap = DepthAiMcap(str(self.path / f"recordings"))
 
         def create_video_file(name):
             if name == 'depth':  # or (name=='color' and 'depth' in self.save):
@@ -56,10 +58,6 @@ class Record:
             # else : fourcc = "GREY" #Simple, single Y plane for monochrome images.
             # files[name] = VideoWriter(str(path / f"{name}.avi"), VideoWriter_fourcc(*fourcc), fps, sizes[name],
             # isColor=name=="color")
-
-        def create_mcap_file(name):
-            mcaps[name] = Mcap(str(self.path / f"{name}"))
-            mcaps[name].imageInit()
 
         while True:
             try:
@@ -78,10 +76,10 @@ class Record:
                         # frames[name].tofile(files[name])
 
                     else:
-                        if name not in mcaps:  # File wasn't created yet
-                            create_mcap_file(name)
+                        if name not in mcap.channels:  # File wasn't created yet
+                            mcap.imageInit(name)
 
-                        mcaps[name].imageSave(frames[name])
+                        mcap.imageSave(frames[name], name)
 
             except KeyboardInterrupt:
                 break
@@ -90,8 +88,7 @@ class Record:
             for name in files:
                 files[name].close()
         else:
-            for name in mcaps:
-                mcaps[name].close()
+            mcap.close()
         print('Exiting store frame thread')
 
     def start(self):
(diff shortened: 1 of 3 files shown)


gen2-record-replay/libraries/depthai_record.py
```diff
@@ -31,43 +39,63 @@ def __init__(self, path: Path, device) -> None:
...
+
+                        # frames[name].tofile(files[name])
+
+                    else:
+                        if name not in mcaps:  # File wasn't created yet
+                            create_mcap_file(name)
 
-                    # if self.rotate != -1: # Doesn't work atm
-                    #     frames[name] = cv2.rotate(frames[name], self.rotate)
+                        mcaps[name].imageSave(frames[name])
```
Erol444, line 84: Can't all streams be saved in a single MCAP file? If that is possible, I assume schemas will get a bit more complicated (as you would need to dynamically create it), but we should go for that.

gen2-record-replay/libraries/depthai_record.py
```diff
@@ -31,43 +39,63 @@ def __init__(self, path: Path, device) -> None:
         calibData.eepromToJsonFile(str(self.path / "calib.json"))
 
         self.convert_mp4 = False
+        self.mcap = mcap
 
     def run(self):
         files = {}
+        mcaps = {}
```
Erol444, line 46: If we have only one .mcap file, this dictionary isn't needed.

gen2-record-replay/libraries/depthai_record.py
```diff
@@ -31,43 +39,63 @@ def __init__(self, path: Path, device) -> None:
...
                 ext = 'h265' if self.quality == EncodingQuality.LOW else 'mjpeg'
                 files[name] = open(str(self.path / f"{name}.{ext}"), 'wb')
             # if name == "color": fourcc = "I420"
             # elif name == "depth": fourcc = "Y16 " # 16-bit uncompressed greyscale image
             # else : fourcc = "GREY" #Simple, single Y plane for monochrome images.
-            # files[name] = VideoWriter(str(path / f"{name}.avi"), VideoWriter_fourcc(*fourcc), fps, sizes[name], isColor=name=="color")
+            # files[name] = VideoWriter(str(path / f"{name}.avi"), VideoWriter_fourcc(*fourcc), fps, sizes[name],
+            # isColor=name=="color")
+
+        def create_mcap_file(name):
```
Erol444, line 60: Dynamically create MCAP schema and create only one .mcap file.

gen2-record-replay/libraries/depthai_record.py
```diff
@@ -242,3 +270,193 @@ def stream_out(name, fps, out, noEnc=False):
         self.pipeline = pipeline
         return pipeline, nodes
 
+
+class Mcap:
```
Erol444, line 274: Move the class to libraries/depthai_mcap.py, and rename it to `DepthAiMcap` (similar to the `DepthAiBags`)

gen2-record-replay/libraries/depthai_record.py
```diff
@@ -242,3 +270,193 @@ def stream_out(name, fps, out, noEnc=False):
...
+        for point in points:
+            buf += struct.pack('f', float(point[0]))
+            buf += struct.pack('f', float(point[1]))
+            buf += struct.pack('f', float(point[2]))
+
+        data = base64.b64encode(buf).decode("ascii")
+
+        tmpTime = time.time_ns()
+        sec = math.trunc(tmpTime / 1e9)
+        nsec = tmpTime - sec
```
Erol444, line 424: Are these variables even used?

gen2-record-replay/libraries/depthai_record.py
```diff
@@ -242,3 +270,193 @@ def stream_out(name, fps, out, noEnc=False):
...
+        # read from .jpeg format to buffer of bytes
+        # byte_im = im_buf_arr.tobytes()
+        byte_im = img.tobytes()
+
+        # data must be encoded in base64
+        data = base64.b64encode(byte_im).decode("ascii")
+
+        tmpTime = time.time_ns()
+        sec = math.trunc(tmpTime / 1e9)
+        nsec = tmpTime - sec
```
Erol444, line 391: Remove.

gen2-record-replay/libraries/depthai_record.py
```diff
@@ -242,3 +270,193 @@ def stream_out(name, fps, out, noEnc=False):
...
+
+        self.writer.add_message(
+            channel_id=self.channel_id,
+            log_time=tmpTime,
```
Erol444, line 395:
```suggestion
            log_time=time.time_ns(),
```

Erol444 (changes requested):

gen2-record-replay/libraries/depthai_record.py
ZigiMigi: yes they are, they are sent in the header of the message

gen2-record-replay/libraries/depthai_record.py
ZigiMigi: Variables are needed.

gen2-record-replay/libraries/depthai_record.py
ZigiMigi: This is needed, as the time must be same as time in header (calling it 2 times produces 2 different times.

Erol444:
TODO: when recording depth and using mcap, use pointCloudSave()

Erol444:
I will be adding depth->pcl conversion and add these commits to the record refactor PR: https://github.com/luxonis/depthai-experiments/pull/369

PR #328, opened 2022-03-20
Created parser for calibration files

ZigiMigi:
Created a parser for calibration files (.json), program is in python format and in jupyter notebook format. Parser reads files and sorts them by camera type. After the data is sorted it can be shown with histograms.


themarpe:
CC: @tersekmatija @VanDavv  @saching13 

Should we integrate this into some tool? 

+ @ZigiMigi (@saching13 ) does it take in account calibration file versioning?

saching13:
> does it take in account calibration file versioning?
It does version 6 only as of now since there are like 10K+ of those alone. We can add version 5 too. But ignore the ones before. 

> Should we integrate this into some tool?

It would be good to have this as in internal tool. But if we streamline all our data as a database. 
Then we can do all this easily using a UI.

tersekmatija:
If we want it publicly available with UI we can integrate it into 'tools' repo that me and @VanDavv are working on for YoloV5 weights export.

Btw, what are the colors on the histogram plot? And can you add some labels to x and y axis + title please? @ZigiMigi

saching13:
What would be more useful to users would be hosting these files on the cloud and have an endpoint to download them if the device id matches.

ZigiMigi:
Loking at the histogram from left to right, its x, y, z. 
And when you are ploting a histrogram from camera 1 and 2 together its x1, x2, y1, y2, z1, z2.
Will integrate the labeling today.

tersekmatija:
> What would be more useful to users would be hosting these files on the cloud and have an endpoint to download them if the device id matches.

So, just hosting static files and downloading them based on device ID? @saching13 

>  And when you are ploting a histrogram from camera 1 and 2 together its x1, x2, y1, y2, z1, z2.

Ah, I see. Makes sense. Maybe using `tab10` for one and `tab20` cmaps for two cameras might keep the colors consistent?
https://matplotlib.org/3.5.1/tutorials/colors/colormaps.html#qualitative @ZigiMigi

ZigiMigi:
Will try it and see if its better.

ZigiMigi:
Is something like this better? (this is tab20 cmap)

Screenshot_1

tersekmatija:
Looking better. Can you also add some labels to x and y axis?

saching13:
> So, just hosting static files and downloading them based on device ID? @saching13

Yes.

parser/parseFiles.py
```diff
@@ -0,0 +1,457 @@
...
+        self.camera0 = dict()
+        self.camera1 = dict()
+        self.camera2 = dict()
+
+        # camera 0 specifications
+        self.camera0["distortionCoeff"] = []
+        self.camera0["rotationMatrix"] = dict()
+        self.camera0["rotationMatrix"]["x"] = []
+        self.camera0["rotationMatrix"]["y"] = []
+        self.camera0["rotationMatrix"]["z"] = []
```
saching13, line 37: It is not XYZ. Change it to `RPY` (Roll, Pitch and Yaw).

parser/parseFiles.py
```diff
@@ -0,0 +1,457 @@
...
+        self.camera0["specTranslation"]["z"] = []
+        self.camera0["translation"] = dict()
+        self.camera0["translation"]["x"] = []
+        self.camera0["translation"]["y"] = []
+        self.camera0["translation"]["z"] = []
+        self.camera0["height"] = []
+        self.camera0["intrinsicMatrix"] = dict()
+        self.camera0["intrinsicMatrix"]["x"] = []
+        self.camera0["intrinsicMatrix"]["y"] = []
+        self.camera0["intrinsicMatrix"]["z"] = []
```
saching13, line 50: What is XYZ here ?

parser/parseFiles.py
```diff
@@ -0,0 +1,457 @@
...
+        self.camera0["specHfovDeg"] = []
+        self.camera0["width"] = []
+
+        # camera 1 specifications
+        self.camera1["distortionCoeff"] = []
+        self.camera1["rotationMatrix"] = dict()
+        self.camera1["rotationMatrix"]["x"] = []
+        self.camera1["rotationMatrix"]["y"] = []
+        self.camera1["rotationMatrix"]["z"] = []
```
saching13, line 59: RPY again

saching13 (changes requested):
can you make this changes to both py and nnoetbook. I will take a look at the updated graphs. 

And as of intrinsics. It should be `f_x`, `f_y`, `c_x` and `c_y`

PR #333, opened 2022-03-29
Added poe tcp streaming demos

Erol444:
TODO: add readmes


themarpe:
Modify such that PoE device is the host and other relevant changes

Erol444:
@themarpe Both options (OAK host or client) are in the folder. What relevant changes?

reallyedutech:
Hi, my name is Ryan. 

i have still not clear how to modify those files. So i have questions.

First: Which one is server? OAK-PoE?
Second: How to put the code on OAK-PoE?

If i am on the wrong way please correct me. 

Best regards,

Erol444:
Hello @reallyedutech ,
- Files can be found [here](https://github.com/luxonis/depthai-experiments/tree/TCP_streaming_poe/gen2-poe-tcp-streaming)
- As you can see on the link above, there are 2 folders; one where OAK is host and one where OAK is the client
- When you run `oak.py`, the pipeline will get uploaded to the (connected) OAK camera. Simultaneously you have to run the `host.py` script, so the TCP communication starts.

themarpe (changes requested):
Add a README and some quick tips on running etc...

Otherwise LGTM

PR #335, opened 2022-04-03
Started refactoring gaze estimation demo. 

Erol444:
https://github.com/luxonis/depthai-experiments/assets/18037362/e132044b-34c8-474a-bb1a-63dec1b5d9a8


Erol444:
Blocker-update: as of 2.19, get/set for seq num is available, but NN result after inference doesn't copy timestamp/seq num from input NNData (only input ImgFrame works).

Erol444:
Gif
ezgif com-optimize

PR #349, opened 2022-05-03
Demo that saves encoded bitstream directly into container

Erol444:
image


Erol444:
TODO: use the `av` library in depthai-record

themarpe (approved):
Cool!

PR #350, opened 2022-05-04
Added depthai-record option to record into mp4 containers

Erol444:


gen2-record-replay/README.md
```diff
@@ -20,12 +20,15 @@ optional arguments:
...
   -q, --quality,      default="HIGH",     Selects the quality of the recording
   -fc,--frame_cnt     default=-1,         Number of frames to record. Record until stopped by default
+  -mp4,--mp4          default=False,      Record directly into mp4 container. Requires `av` library
 ```
 
 For the `frame_cnt`, -1 means it will record streams until user terminates the program (`CTRL+C`). If you select eg. `-fc 300 --fps 30`, recording will be of 300 frames (of each stream), for a total of 10 seconds.
 
 `quality` specifies the quality of encoded video streams. It can either be `BEST` (lossless encoding), `HIGH`, `MEDIUM` or `LOW`. More information regarding **file sizes and quality of recordings** can be [found here](encoding_quality/README.md).
 
+`-mp4` argument will enable recording streams directly into `.mp4` videos so you can view them with standard video player. To enable need, make sure you have [PyAV](https://github.com/PyAV-Org/PyAV) installed (`python3 -mpip install av`). More [info here](../gen2-container-encoding).
```
themarpe, line 30: Make it default + add this under `requirement.txt`. 

In the experiment, try to import this dependency and warn in case of fallback (back to containerless saving format)

themarpe (changes requested):

themarpe (changes requested):
IMO Don't disable the option to save raw streams, just make the `-mp4` option default on.

themarpe (approved):

PR #352, opened 2022-05-10
Add CREStereo experiment

N950:


tersekmatija (changes requested):
Let's add blobs to our shared cloud and write a simple download.py script that download them and places them into models directory. Something like [this](https://github.com/luxonis/depthai-experiments/blob/master/gen2-yolo/host-decoding/download.py).

tersekmatija:
Can we add some warning (maybe even to the README), that blob should match the correct shape? @N950

tersekmatija (approved):
Thanks, looks good!

PR #372, opened 2022-06-27
Change version to 2.16.0.0

moratom:
Updated library versions and fixed some examples to work with 2.16.0.0 depthai-python version.

Now all examples use the new version, except for the **gen2-lossless-zooming** experiment:

- The experiment is currently on depthai version - 2.13.2.0.dev0+87552f and it works.
- The experiment works on the branch image_manip_refactor (b94176c)
- It doesn't work on the default 2.16.0.0 version though - I kept the old version in the requirements.txt for now.


gen2-crestereo-stereo-matching/requirements.txt
```diff
@@ -1,4 +1,4 @@
-opencv-python==4.5.1.48
-depthai==2.15.2
+opencv-python==4.5.5.62
```
Erol444, line 1: Is this specifically needed for this demo? If so, please add a comment that this version is required

Erol444 (approved):
LGTM and thanks!

PR #379, opened 2022-07-16
Fix multi-stage NN demos

Erol444:
TODO:
- [X] update other examples that use multi-stage NN pipeline architecture


Erol444:
LGTM thanks!

PR #385, opened 2022-08-08
Demo program to save point cloud and rectified image data.

marco-paladini:
Moving the https://github.com/luxonis/depthai-python/pull/655 pull request to the experiments repo.

This PR provides users with a debugging program that computes a point cloud and writes all debugging data to disk.
Also added a minimal pcd viewer and a README for the demo program.

### DIFF
--- a/gen2-pointcloud/save_point_cloud/main.py
+++ b/gen2-pointcloud/save_point_cloud/main.py
@@ -1,0 +1,225 @@
+"""Display depth+rgb aligned and save to point cloud file.
+Adapted from https://github.com/marco-paladini/depthai-python/blob/main/examples/StereoDepth/rgb_depth_aligned.py"""
+#!/usr/bin/env python3
+
+import json
+import time
+
+import cv2
+import depthai as dai
+import numpy as np
+import open3d as o3d
+
+# Weights to use when blending depth/rgb image (should equal 1.0)
+rgbWeight = 0.4
+depthWeight = 0.6
+
+
+def updateBlendWeights(percent_rgb):
+    """
+    Update the rgb and depth weights used to blend depth/rgb image
+
+    @param[in] percent_rgb The rgb weight expressed as a percentage (0..100)
+    """
+    global depthWeight
+    global rgbWeight
+    rgbWeight = float(percent_rgb) / 100.0
+    depthWeight = 1.0 - rgbWeight
+
+
+# Optional. If set (True), the ColorCamera is downscaled from 1080p to 720p.
+# Otherwise (False), the aligned depth is automatically upscaled to 1080p
+downscaleColor = True
+fps = 30
+# The disparity is computed at this resolution, then upscaled to RGB resolution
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon:
Thank you!

marco-paladini:
This demo program was based off rgb_depth_aligned but now I noticed the `rgbd-pointcloud` program in this repo does mostly everything that is done in `save_point_cloud` except for saving the data to disk, so potentially this PR should be closed and the save code added to `rgbd-pointcloud`?

gen2-pointcloud/save_point_cloud/main.py
```diff
@@ -0,0 +1,232 @@
...
+            # Need to have both frames in BGR format before blending
+            if len(frameDepth.shape) < 3:
+                frameDepth = cv2.cvtColor(frameDepth, cv2.COLOR_GRAY2BGR)
+            blended = cv2.addWeighted(frameRgb, rgbWeight, frameDepth, depthWeight, 0)
+            cv2.imshow(blendedWindowName, blended)
+            frameRgb = None
+            frameDepth = None
+
+        if cv2.waitKey(1) == ord("q"):
+            # save all images
```
Erol444, line 186: maybe we could save images when pressing `s` as save instead?

gen2-pointcloud/save_point_cloud/main.py
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        queueEvents = device.getQueueEvents(
+            ("rgb", "depth", "rectifiedLeft", "rectifiedRight")
+        )
+        for queueName in queueEvents:
+            packets = device.getOutputQueue(queueName).tryGetAll()
+            if len(packets) > 0:
+                latestPacket[queueName] = packets[-1]
+
```
Erol444, line 153: This logic of getting frames isn't ideal. Maybe jut use get() instead, or maybe add syncing logic to it (based on sequence number)?

gen2-pointcloud/save_point_cloud/main.py
```diff
@@ -0,0 +1,232 @@
...
+# rectified stereo
+xoutRectifLeft = pipeline.create(dai.node.XLinkOut)
+xoutRectifRight = pipeline.create(dai.node.XLinkOut)
+xoutRectifLeft.setStreamName("rectifiedLeft")
+xoutRectifRight.setStreamName("rectifiedRight")
+stereo.rectifiedLeft.link(xoutRectifLeft.input)
+stereo.rectifiedRight.link(xoutRectifRight.input)
+
+# Connect to device and start pipeline
+with device:
```
Erol444, line 109: since you are using context manager, move `with dai.Device(pipeline) as device` here (like in all other demos)

Erol444 (changes requested):

Erol444:
@marco-paladini yes I agree I would just expand the existing demo to add saving option when pressing eg. `s` key. Thanks!

PR #387, opened 2022-08-14
update readme

kevinrigot:


Erol444 (approved):

Erol444 (changes requested):
That's actually not the case, see argument [parser here](https://github.com/luxonis/depthai-experiments/blob/master/gen2-record-replay/record.py#L30).

PR #399, opened 2022-09-07
added reverse transformation from 3D to 2D

klemen1999:
Added reverse transformation with visualization of 3D bounding box points to 2D image plane.

### DIFF
--- a/gen2-box_measurement/box_estimator.py
+++ b/gen2-box_measurement/box_estimator.py
@@ -28,6 +28,7 @@ class BoxEstimator():
 
         self.bounding_box = None
         self.rotation_matrix = None
+        self.translate_vector = None
 
         self.max_distance = max_distance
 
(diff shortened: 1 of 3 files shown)


miknai:
Had to set p1 and p2 as tuples like below. Otherwise, the program threw the SystemError: new style getargs format but argument is not a tuple
```
p1 = tuple(int(x) for x in img_points[line[0]][0])
p2 = tuple(int(x) for x in img_points[line[1]][0])
```
Now, I can run the code example, but it does not how me the color and point cloud images. The program gets frozen as soon as I run it.

klemen1999:
The error you got was due to older opencv version. I updated the requirements.txt and not it should be fixed. 
As for the freezing, does the app run correctly if you comment out 2D visualization in main.py (lines 159 and 160)? I would need some more information to try to reproduce this.

miknai:
I see. The packages are now updated based on the requirements.txt.

I still face the issue **with and without** the two lines (159 and 160 in main.py)

miknai:
This is what I get... It freezes every time. Sometimes the color window pops up, but not the point cloud window.

Screenshot from 2022-09-08 10-35-07

Screenshot from 2022-09-08 10-37-42

miknai:
I use OAK-D-POE-PRO and the below is the versions of FW and depthai

Screenshot from 2022-09-08 10-49-52

miknai:
so, I tried the code with other cameras (OAK-D-POE) and it worked fine. The issue was my OAK-D-POE-PRO camera... have to do factory reset maybe.

miknai:
I have three cameras (1 OAK-D-POE-PRO and 2 OAK-D-POE) and now only 1 OAK-D-POE camera works with this code. Very confusing.

klemen1999:
Based on the camera stream I would say that the problem is the setting in which you are testing the code. The current approach has some limitations and the main one is that the region of interest has to consist only of ground (a plane) and a box. Check the images in [README.md](https://github.com/luxonis/depthai-experiments/blob/gen2-box-measurement-add_visualisation/gen2-box_measurement/README.md)  to get an idea of what is the assumed box position by the algorithem and other limitations of the approach.

miknai:
I figured out my issue. After doing "pip3 install -r requirements.txt" with the requirement file in box measurement dir, the latest version of depthai (depthai==2.17.3.0) gets overwritten by the old one, depthai==2.16.0.0 which is specified in the file. With the certain version of depthai and the latest FW, it gave me difficult time to use the stereo sensors. 

After upgrading the depthai version to the latest, I was able to avoid the issues I wrote above. It would be a good idea to update the depthai version in requirements.txt

miknai:
Other than that, I don't have any issues. Thank you for you effort implementing the feature very quickly. Much appreciate it.

moratom:
Thanks for the feedback @miknai, I updated the requirements.

PR #405, opened 2022-09-20
fixed gen2-multiple-devices/main.py 

Erol444:
OAK cameras now connect in parallel (using threads), which leads to faster bootup time and potentially resolve the issue with connecting oak poe cameras


Erol444:
LGTM

PR #409, opened 2022-10-11
Box measurement point cloud fusion

njezersek:
Applied box measurement algorithm (https://github.com/luxonis/depthai-experiments/tree/master/gen2-box_measurement) on point clouds from multiple cameras merged together.


Erol444 (approved):
Not sure if all the `sample_data` files are required, otherwise lgtm. Thanks!

PR #415, opened 2022-10-26
Z-accuracy and spatial noise

njezersek:
Proof of concept for testing z-accuracy and spatial noise.


saching13:
@njezersek  can you remove binary files and json files.
Add them to .gitignore in the sub folder

alex-luxonis:
To not introduce the binary files into the repo history (that would increase its size), rewrite the branch history to remove the file additions, or, easier, use 'squash' when merging:
image

saching13:
@njezersek  Can you make a small video of usage and share it here ?

depth-test_z-acc_spatial-noise/oak_camera.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def _create_pipeline(self):
+        pipeline = dai.Pipeline()
+
+        # Depth cam -> 'depth'
+        mono_left = pipeline.createMonoCamera()
+        mono_right = pipeline.createMonoCamera()
+        mono_left.setResolution(dai.MonoCameraProperties.SensorResolution.THE_400_P)
+        mono_left.setBoardSocket(dai.CameraBoardSocket.LEFT)
+        mono_right.setResolution(dai.MonoCameraProperties.SensorResolution.THE_400_P)
```
saching13, line 60: make the resolution arguments.

saching13 (changes requested):

saching13:
I haven't checked the math logic yet. Will do it after the following updates.
In addition to above comments, it looks like you are testing two devices individually. 

What we need to do is modify it to use the results from other device as ground truth. 
1. Create the Base truth Wall to adjust the angle of the devices. 
2. Create the Ground Turth Plane using the secondary device. 
3. Use that ground Truth to estimate the Z-accuracy. 
4. Fill rate, Plane FIT RMSE and Sub-Pixel RMSE stays same since they are Spatial Noise Estimators.

depth-test_z-acc_spatial-noise/oak_camera.py
njezersek: Done

depth-test_z-acc_spatial-noise/oak_camera.py
```diff
@@ -0,0 +1,125 @@
...
+        mono_right.setBoardSocket(dai.CameraBoardSocket.RIGHT)
+        cam_stereo = pipeline.createStereoDepth()
+        cam_stereo.setDefaultProfilePreset(dai.node.StereoDepth.PresetMode.HIGH_DENSITY)
+        cam_stereo.initialConfig.setMedianFilter(config.median)
+        cam_stereo.initialConfig.setConfidenceThreshold(config.confidence_threshold)
+        cam_stereo.setLeftRightCheck(config.lrcheck)
+        cam_stereo.setExtendedDisparity(config.extended)
+        cam_stereo.setSubpixel(config.subpixel)
```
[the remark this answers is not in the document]
njezersek: Done

depth-test_z-acc_spatial-noise/astra_camera.py
```diff
@@ -0,0 +1,82 @@
...
+    def __del__(self):
+        self.image_stream.stop()
+        self.depth_stream.stop()
+
+    def _load_calibration(self):
+        # Intrinsics (https://3dclub.orbbec3d.com/t/access-intrinsic-camera-parameters/2784)
+        hfov = self.image_stream.get_horizontal_fov()
+        vfov = self.image_stream.get_vertical_fov()
+
+        intrinsics = np.array([
```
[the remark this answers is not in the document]
njezersek: Yes, it does change and no they don't have an option to get it (at least I couldn't find it). You can get the camera FOV though and use it to compute the intrinsic matrix (as described here https://3dclub.orbbec3d.com/t/access-intrinsic-camera-parameters/2784). But I got better results by manually calibrating the camera to get the matrix.

I will make the intrinsic matrix an external parameter and if not provided compute the matrix from the FOV.

depth-test_z-acc_spatial-noise/config.py
```diff
@@ -0,0 +1,20 @@
...
+min_range = 100 			# mm
+max_range = 2000			# mm
+
+# Median filter
+# Options: MEDIAN_OFF, KERNEL_3x3, KERNEL_5x5, KERNEL_7x7
+median   = dai.StereoDepthProperties.MedianFilter.KERNEL_7x7
+
+
+# CAMERA POSITION (ground truth)
+camera_wall_distance = 0.996 # m
```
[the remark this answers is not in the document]
njezersek: Done

njezersek:
Here is a screen recording demonstrating the usage (I had some problems with the screen recorder, so I recorded the screen with my phone)

Steps taken:
- Select ROI on both cameras
- Selected the Astra cam with the `2` key
- Set ground truth from the Astra cam measurement (`g` key)
- Selected the OAK cam with the `1` key
- Fit the plane to the OAK's point cloud (`f` key)
- Run the tests (`t` key)


https://user-images.githubusercontent.com/10494687/202206342-58c7b864-ae3b-4a47-a0d7-d81f4511385e.mp4

saching13:
> * Set ground truth from the Astra cam measurement (`g` key)

is this done using avg of the region selected ?

depth-test_z-acc_spatial-noise/depth_test.py
```diff
@@ -0,0 +1,174 @@
...
+		self.fitted = False
+
+		self.z_accuracy_medians = []
+		self.z_means = []
+		self.spatial_noise_rmses = []
+		self.samples = 0
+
+	def set_ground_truth(self, point_cloud: o3d.geometry.PointCloud):
+		points = np.asarray(point_cloud.points)
+		G = np.ones_like(points)
```
saching13, line 27: why one's like ?

depth-test_z-acc_spatial-noise/depth_test.py
```diff
@@ -0,0 +1,174 @@
...
+			f"Vertical tilt: {abs(vertical_tilt) * 180/np.pi}° {'UP' if vertical_tilt < 0 else 'DOWN'}"
+		)
+
+	def project_on_plane(self, point_cloud: o3d.geometry.PointCloud):
+		points = np.asarray(point_cloud.points)
+		G = np.ones_like(points)
+		G[:, 0] = points[:, 0]
+		G[:, 1] = points[:, 1]
+		plane_fit = o3d.geometry.PointCloud(o3d.utility.Vector3dVector(np.hstack([points[:,0:2], G @ np.matrix(self.plane_coeffs).T])))
+		return plane_fit
```
saching13, line 72: Why is Z value not used in projection here ? Can you add an explanation here ?

saching13: Are you checking like if Points in a Plane at 1 unit distance intersects with the tilted Created coeffs ?

saching13: and those values are used for compute_z_accuracy ? Doesn't add up.

depth-test_z-acc_spatial-noise/depth_test.py
njezersek: The `G` matrix is not even used here - apparently I copied too much. I will remove it.

depth-test_z-acc_spatial-noise/depth_test.py
njezersek: > and those values are used for compute_z_accuracy ? Doesn't add up.

These values are not used for computing the z-accuracy. Z-accuracy is computed from `point_cloud_corrected` which is just rotated original point cloud. [here](https://github.com/njezersek/depthai-experiments/blob/5fc869407c735c70001f059bd75852635d97866a/depth-test_z-acc_spatial-noise/depth_test.py#L92=)

It is used for computing the spatial noise through. But now when I think about it I think it is unnecessary. Since the projected points get rotated the z components shuld be the same as `self.camera_wall_distance`. 

I made this function primarly to visualise the fitted plane and sanity check the code.

> Are you checking like if Points in a Plane at 1 unit distance intersects with the tilted Created coeffs in fit_plane ?

I'm not sure what you mean by that but probably not.

> Why is Z value not used in projection here ? Can you add an explanation here ?

I made a projection along the z axis. Perhaps it would be better to make a perpendicular projection to the fitted plane. But again I don't think it matters since only the z component is used for computing the spatial noise.

saching13, line 72: can you elaborate on what plane_fit variable contains here ? it looks like pointCloud.

depth-test_z-acc_spatial-noise/depth_test.py
```diff
@@ -0,0 +1,174 @@
...
+		normal = np.array([k_x, k_y, -1])
+		normal = normal / np.linalg.norm(normal)
+
+		self.plane_normal = normal
+		self.plane_distance = k_z
+		self.plane_coeffs = (k_x, k_y, k_z)
+
+		self.fitted = True
+
+		return normal, k_z
```
saching13, line 51: This fit is happening on OAK's PointCloud or Astra ?

depth-test_z-acc_spatial-noise/depth_test.py
```diff
@@ -0,0 +1,174 @@
...
+		self.samples += 1
+
+		return spatial_noise, z_accuracy
+
+	def compute_spatial_noise(self, point_cloud_corrected: o3d.geometry.PointCloud, plane_fit_corrected: o3d.geometry.PointCloud):
+		plane_fit_corrected_points = np.asarray(plane_fit_corrected.points)
+		point_cloud_corrected_points = np.asarray(point_cloud_corrected.points)
+
+		z_error = plane_fit_corrected_points[:,2] - point_cloud_corrected_points[:,2]
+		z_error = np.sort(z_error)
```
saching13, line 105: Use disparity here. Considering plane estimatin is correct. 
image

depth-test_z-acc_spatial-noise/depth_test.py
njezersek: Yes it is a point cloud containing the projections along z-axis on the fitted plane.

depth-test_z-acc_spatial-noise/depth_test.py
njezersek: Depends on which camera is selected when you press the `f` key.

If we can ensure the cameras are perfectly parallel it wuld be preferable to use Astra for this. Otherwise I think it would be better to fit a plane on the OAK's point cloud to avoid possible missalignment errors.

depth-test_z-acc_spatial-noise/depth_test.py
njezersek: I will have to take a deeper look at that. Right now all calculations are based on the point cloud. 

Isn't the z value directly proportional to disparity? Why not use just the z value of the points in the point cloud since they are already in the correct units (meters instead of pixels)?

saching13: > Isn't the z value directly proportional to disparity?
it is inversely proportional. 
Might be fine. but would be easier to do a comparison was what I was thinking.

depth-test_z-acc_spatial-noise/depth_test.py
```diff
@@ -0,0 +1,174 @@
...
+		self.samples = 0
+
+	def set_ground_truth(self, point_cloud: o3d.geometry.PointCloud):
+		points = np.asarray(point_cloud.points)
+		G = np.ones_like(points)
+		G[:, 0] = points[:, 0]
+		G[:, 1] = points[:, 1]
+		Z = points[:, 2]
+
+		self.camera_wall_distance = -np.mean(Z)
```
saching13, line 32: Can you cross check if mean or median would be best here by some trial and errors ?

saching13: I have a feeling median might do better.

njezersek: It is hard to tell which is better but the median appears more consistent over multiple measurements so I'll change it to median. The difference between mean and meadian is less than a millimeter at one meter distance.

saching13 (approved):
LGTM. Thanks

PR #417, opened 2022-10-27
Kalman filter

vitakomel:


gen2-bounding-box-filtering/kalman/README.md
```diff
@@ -0,0 +1,6 @@
+# Kalman filter
```
Erol444, line 1: Maybe expand this a tiny bit more - like why is it useful, and perhaps some links to relevant kalman filter tutorials (eg. YT)

gen2-bounding-box-filtering/kalman/errors.csv
```diff
@@ -0,0 +1,20 @@
+﻿Ground truth,OAK-D Error [%]
```
Erol444, line 1: Maybe we could just use theoretical depth error for this estimation ([here](https://docs.google.com/spreadsheets/d/1ymn-0D4HcCbzYP-iPycj_PIdSwmrLenlGryuZDyA4rQ/edit#gid=0))

gen2-bounding-box-filtering/kalman/kalman_filter.py
```diff
@@ -0,0 +1,218 @@
+import depthai as dai
+import blobconverter
+import cv2
+import time
+import numpy as np
+import pandas as pd
+import imageio
```
Erol444, line 7: is this required?

gen2-bounding-box-filtering/kalman/kalman_filter.py
```diff
@@ -0,0 +1,218 @@
+import depthai as dai
+import blobconverter
+import cv2
+import time
+import numpy as np
+import pandas as pd
```
Erol444, line 6: if we use formula above (isntead of parsing csv) we can also remove pandas dependency

Erol444 (changes requested):
Why is this folder inside `gen2-bounding-box-filtering`?

Erol444 (approved):

PR #420, opened 2022-10-31
Seven Segment experiment init

N950:
Added seven segment demo

### DIFF
--- a/gen2-seven-segment-recognition/download.py
+++ b/gen2-seven-segment-recognition/download.py
@@ -1,5 +1,0 @@
-import gdown
-
-if __name__ == "__main__":
-
-    gdown.download("https://drive.google.com/uc?id=1LJG6xBnJvhHrVkNgsW1C7xaUGGD7totT", output="blobs/7_SEG_256x750_MONO_simplified_openvino_2021.4_6shave.blob")
--- a/gen2-seven-segment-recognition/download.py
+++ b/gen2-seven-segment-recognition/download.py
@@ -1,0 +1,5 @@
+import gdown
+
+if __name__ == "__main__":
+
+    gdown.download("https://drive.google.com/uc?id=1LJG6xBnJvhHrVkNgsW1C7xaUGGD7totT", output="blobs/7_SEG_256x750_MONO_simplified_openvino_2021.4_6shave.blob")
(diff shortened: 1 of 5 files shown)


gen2-seven-segment-recognition/download.py
```diff
@@ -0,0 +1,5 @@
+import gdown
+
+if __name__ == "__main__":
+
+    gdown.download("https://drive.google.com/uc?id=1LJG6xBnJvhHrVkNgsW1C7xaUGGD7totT", output="blobs/7_SEG_256x750_MONO_simplified_openvino_2021.4_6shave.blob")
```
Erol444, line 5: Could we rather add the model to depthai-model-zoo, and use blobconverter to download it?

gen2-seven-segment-recognition/main.py
```diff
@@ -0,0 +1,97 @@
...
+
+# MonoCamera
+monoRight = pipeline.create(dai.node.MonoCamera)
+monoRight.setBoardSocket(dai.CameraBoardSocket.RIGHT)
+monoRight.setResolution(dai.MonoCameraProperties.SensorResolution.THE_720_P)
+
+# Crop manip
+manip1 = pipeline.create(dai.node.ImageManip)
+manip1.initialConfig.setCropRect(0, 0, 0.694444444, 0.355555556)
+monoRight.out.link(manip1.inputImage)
```
Erol444, line 38: Why the cropping?

Erol444 (approved):

gen2-seven-segment-recognition/main.py
N950: @Erol444  That's the ROI as the model was trained on 256x750 with the 7 seg display filling the majority of it, resizing the whole frame will change the aspect ratio it was trained on and the 7 seg display will be a small part of it
This is recognition only without detection

Erol444: @N950 so to preserve aspect ratio, wouldn't it be more efficient to just set the `resizeManip.setKeepAspectRatio(True)`, instead of having a cropping manip?

N950: @Erol444 Aspect ratio and also ROI
`resizeManip.setKeepAspectRatio(True)` would still take a larger part of the frame
Here is what I used for reference on this

Left to right: full frame | setKeepAspectRatio | crop + resize
aspect_ratio

Erol444:
image

PR #450, opened 2023-01-25
Updated IMU/depth/rgb syncing demo so it's better fit for VIO solutions

Erol444:
Merge together with https://github.com/luxonis/depthai-docs-website/pull/347


szabi-luxonis (approved):
Thanks

PR #463, opened 2023-03-09
Moved API implementation to its own folder, updated SDK example a bit

Erol444:


daniilpastukhov:
Perhaps it'd be better to merge into `refactor` branch. Also do you think that we shouldn't use `sdk` folder? In `refactor` branch every folder contains api/sdk folders with the corresponding implementation

PR #474, opened 2023-04-29
Bugfix: Height property to get the height param

msanterre:
Small bugfix to accurately fetch the height parameter from the request instead of the width within the `height` property.


Erol444 (approved):

Erol444:
Look good, thank you @msanterre !

PR #480, opened 2023-05-20
Updated both record and replay scripts. 

Erol444:
**Requires 1.11 SDK release**

### DIFF
--- a/gen2-license-plate-recognition/main.py
+++ b/gen2-license-plate-recognition/main.py
@@ -6,7 +6,7 @@ import blobconverter
 import cv2
 import depthai as dai
 import numpy as np
-from depthai_sdk import FPSHandler
+from depthai_sdk.fps import FPSHandler
 
 parser = argparse.ArgumentParser()
 parser.add_argument('-nd', '--no-debug', action="store_true", help="Prevent debug output")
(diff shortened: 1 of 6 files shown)


Jalen-Xu:
Hello,

I use the latest version of record.py file to record depth data, but it still prompts an error, as shown below.

企业微信截图_16884639141778

Erol444:
Hi @Jalen-Xu, for depth, you'd need to record either in MCAP or ROSBAG, not VIDEO.

Jalen-Xu:
Hi,

Thank you for your support!
By executing the command "python record.py -save depth -type ROSBAG" , I saved a depth image (in rosbag format). Then I ran the RealSense Viewer software and tried to replay this depth image, but I failed. It prompted some errors, as shown below.
企业微信截图_16885401575893

Erol444:
@Jalen-Xu  Could you try to open rosbag with foxglove studio instead?

Jalen-Xu:
Thank you. 
I was able to successfully open the rosbag depth image file through the foxglove studio software. But I can't see the x/y/z values of each pixel point. Compared to RealSense Viewer software, foxglove studio software is inconvenient.

Erol444:
Hi @Jalen-Xu ,
What about recording pointcloud instead of the depth frame?

PR #498, opened 2023-09-08
Fix two bugs in imu-frames-timestamp-sync

sgraybi:
Fix an import bug from depthai_sdk.
td2ms only takes one argument.


Erol444 (approved):

Erol444:
Thanks for the PR!

PR #530, opened 2024-08-21
Update RTSP stream example

Erol444:
Tested on Mac M2 14, Win 10
Doesn't work on Ubuntu 22, have yet to debug that
It relies on this cross-platform media server: https://github.com/bluenviron/mediamtx


Erol444:
Works on Ubuntu 24.04 (ffmpeg 6.1.1), but I wasn't able to get it working on Ubuntu 22.04 (ffmpeg 4.2.2, which threw many problems)

PR #555, opened 2024-11-29
Segmentation models example

aljazkonec1:
This PR adds a general sicript for running all pure segmentation models that is, models that return the message type `dai_nodes.SegmentationMask`, otherwise an exception is raised. The main addition is the annotation node that blends the segmentation mask over the original image frame.


klemen1999 (approved):
LGTM

PR #557, opened 2024-12-05
feat: re-organize neural networks section

jkbmrz:
This PR introduces organizational changes to the neural networks section. The main aim is to:
- simplify utilization for the user,
- give a more structured overview of the available content,
- remove unnecessary / repeating / obsolete content.

TODOs:
- remove the remaining advanced examples that are obsolete (aka. can be substituted by the `basic-example`) and list all remaining examples into a table;
- prepare media for each of the listed HubAI models (e.g. an image / gif / video file that the user can use to test out the model, and output media file so that the user can quickly get a better understanding of the model performance without the need to run the model),
- think about renaming the section from `neural-networks` to `ai-models` / `ai-inference` (following the same naming convention as in the docs).
- make sure all the experiments (basic and advanced ones) are visible enough to the user! Either have advanced experiments in a separate folder to make the basic example stick out mode, or in the same folder, to make the advanced examples stick out more. This could also be solved by good structure of the readme file


jkbmrz:
There is still some things to fix/add/remove (see the TODOs above) but I'm opening this PR to get some general feedback based on which I'll be able to finish the refactor.

PR #566, opened 2024-12-19
ref: Gen3 Restructure

dominik737:
Removing gen2 folder. Moving experiments from gen3 folder to root folder.


klemen1999:
We'll probably want to define a nicer landing page README. We have quite a nice structure in neural-networks already but it might make sense to expand on it and put it as the main README (maybe still keep more detailed version in nn folder but have all important information also in the main one). Can maybe sync offline on the best structure for it

dominik737:
Agree. I've got feeling from conversation with @tersekmatija that the new structure should in place ASAP. So I made made the PR as soon as it was possible. But the changes to the README would be nice.

We could do it such that I load the table(s) from the Neural Networks README and adjust the rest to have the same style. Also I think it would be also nice to link the README in the landing page, as it describes the Generic/Advanced example.

klemen1999 (approved):

PR #576, opened 2025-01-08
Gen3: Animal pose estimation.

kkeroo:
This PR adds 2-stage Animal detection and pose estimation. YOLOv6 is used for detection and SuperAnimal for pose estimation. The pipeline works on both platforms (RVC2 and RVC4). There are some performance issues with oak visualizer so RVC4 performance lower. I checked with cv2 visualization and it is realtime so no problems in the pipeline.

We add some padding when cropping detections so the pose estimation model has better input image.

Readme, demo GIF and standalone app is also included.


neural-networks/advanced-examples/pose-estimation/animal-pose/main.py
```diff
@@ -0,0 +1,109 @@
...
+
+    script = pipeline.create(dai.node.Script)
+    detection_nn.out.link(script.inputs['det_in'])
+    detection_nn.passthrough.link(script.inputs['preview'])
+    script_filename = "script_rvc2.py" if platform == "RVC2" else "script_rvc4.py"
+    script.setScriptPath(Path(__file__).parent / script_filename)
+
+    if platform == "RVC2":
+        pose_manip = pipeline.create(dai.node.ImageManip)
+        pose_manip.initialConfig.setResize(256, 256)
```
klemen1999, line 75: Let's get this values dynamically from the pose_nn_archive so it is not hardcoded (same for pose_manip under RVC4)

neural-networks/advanced-examples/pose-estimation/animal-pose/oak_standalone.toml
```diff
@@ -0,0 +1,12 @@
+identifier = "com.luxonis.xfeat-example"
```
klemen1999, line 1: Update to correct identifier

neural-networks/advanced-examples/pose-estimation/animal-pose/media/cow-walking.gif
klemen1999, line 1: Is it possible to not connect tail keypoint to anything? Seems like now it's connected to some keypoint on the head?

klemen1999 (approved):
LGTM, left some comments

neural-networks/advanced-examples/pose-estimation/animal-pose/main.py
kkeroo: Fixed in [f801120](https://github.com/luxonis/depthai-experiments/pull/576/commits/f801120dac7e196c4e84fd6531e45fa92830b208).

neural-networks/advanced-examples/pose-estimation/animal-pose/oak_standalone.toml
kkeroo: Thanks, fixed in [74dd29a](https://github.com/luxonis/depthai-experiments/pull/576/commits/74dd29a618313828562bdc00df137dddbd6e7412).

neural-networks/advanced-examples/pose-estimation/animal-pose/media/cow-walking.gif
kkeroo: I think the problem is with wrongly detected keypoint. I checked the drawing rules and tail should be connected to the head. I manually disabled some keypoints so the visualization is better. New GIF in [25a6a7c](https://github.com/luxonis/depthai-experiments/pull/576/commits/25a6a7cb65f316520e782599090c9766764a7dc5).

klemen1999: > tail should be connected to the head

Should or shouldn't? I would say shouldn't?

kkeroo: Sorry, I meant shouldnt. So, the problem is not in wrong visualization rules but in badly detected keypoint because we dont filter keypoints based on confidence but rather draw all of them even if the confidence is 0.01.

Should I filter out low confidences and draw only part of the skeleton?

PR #580, opened 2025-01-10
NN readme update, added pre-commit and CI

klemen1999:
- Refactored the NN README to now group experiments by task
  - Also reorganized the tables so now Gen3 example is the main field, HubAI model(s) is more clearly linked to example and Gen2 example is at the end because RVC2 and RVC4 fields are only relevant to Gen3
- Removed "advanced-examples" directory so user can more quickly get to specific experiment
  - Generic example is still its own folder, outside of any specific task
- Added pre-commit
- Added CI that triggers on `gen3` PRs and runs pre-commit

TODO:
- Add some visualizations (either from generic or "advanced" examples) to each task in main README so user knows what type of experiments you can expect under this folder
  - Previously we had "Illustrations" for each experiment but I think having just 2/3 most interesting ones actually showing would be better. We still want visualizations in every advanced example folder in its specific README though, this is just regarding the main README.
- Update links on HubAI when PR gets merged

Tip: Because of reorganizing and pre-commits causing changes to lots of files I would suggest you rather review the PR as a whole branch, not look at the exact changes since there aren't any (except pre-commit stuff). Or look at specific commits (mainly the first one).


dominik737:
I was thinking whether to rename and move the `generic-example` to apps instead. 

It is quite robust and does a lot of things. I wouldn't say it's single purpose "experiment". With the change we will have under neural-networks section only the advanced examples now, so it would fit the rule, that in the folder level there are advanced examples (mostly 2-stage nns) only, not a generic "unicorn".
If we decide to move the example to apps folder I would propose we change the name to something like `neural-networks-visualization`.

Thoughts?

klemen1999:
Why I am reluctant to do this is because one of the goals of the neural-networks folder is to also be like a map between gen2 NN experiments and how are these now done in gen3 (example from users persective: I was doing yolo detection with DAIv2, how do I achieve now the same with gen3). And generic example covers a lot of these simple examples now so IMO it makes sense to keep it here with all other NN examples.

jkbmrz:
The restructure looks good! I'd still make the generic-experiment more visible though. Maybe by setting it’s name to caps-lock or to `_generic_example` so that it would show on the top of the page?

Also, the Overview section in the README needs to change slightly as there is no "advanced examples" anymore.

jkbmrz:
> I was thinking whether to rename and move the generic-example to apps folder instead.

@dominik737 what's the added value of the apps folder? If it's meant to showcase the more general examples for the whole repository, then I think it makes sense to move the generic-example there (or even better, in my opinion, to add a symbolic link to it and the generic-example stays in the nn folder).

aljazkonec1:
Looks good, I agree with @jkbmrz regarding README and visibility of generic-example. In a later PR we should add more gifs showing the usage.

aljazkonec1 (approved):

dominik737:
> what's the added value of the apps folder? If it's meant to showcase the more general examples for the whole repository, then I think it makes sense to move the generic-example there (or even better, in my opinion, to add a symbolic link to it and the generic-example stays in the nn folder).

@jkbmrz the purpose is basically to put there apps that we want to have open-sourced. Right now there is only the default app, but I believe there are more to come. I would not put those apps into NNs folder. For example the default app uses the yolo network for detections, but it's purpose is not only to show our NN's, but to also show our depth, pointcloud etc. Thus

dominik737:
> The restructure looks good! I'd still make the generic-experiment more visible though. Maybe by setting it’s name to caps-lock or to _generic-example so that it would show on the top of the page?

If we can make it first in the list it would be nice. But best if we can figure any other symbol than underscore, which is used in Python for protected/private variables, classes and so on.

PR #592, opened 2025-01-20
Gen3: Human-machine-safety 

kkeroo:
This PR  updates Human machine safety experiment:

- YOLOv6 nano dangerous object detector (cup, bottle) and MediaPipe palm detector
- RVC2 & RVC4
- Depth
- Measure object distance and print warning if too close
- Standalone
- GIF


neural-networks/object-detection/human-machine-safety/main.py
```diff
@@ -97,15 +145,22 @@
...
+    annotation_node = pipeline.create(AnnotationNode)
+    detection_filter.output.link(annotation_node.detections_input)
+    measure_object_distance.output.link(annotation_node.distances_input)
+
+    visualizer.addTopic("Color", camera_output)
+    visualizer.addTopic("Detections", annotation_node.out_detections)
     visualizer.addTopic("Distances", visualize_distances.output)
     visualizer.addTopic("Alert", show_alert.output)
-    visualizer.addTopic("Color", color_out)
+
```
klemen1999, line 156: Would be nice to also include "disparity" to visualizer and show as separate stream also on GIF - it looks more interesting IMO. But perhaps it would need to go through a seprate "visualizer node" before just so we apply a colormap to make it look nicer.

kkeroo:
@klemen1999 I added GIF to the general table under the object detection, so we have now two GIFs in that section.

neural-networks/object-detection/human-machine-safety/main.py
```diff
@@ -1,79 +1,130 @@
...
+        ),
         presetMode=dai.node.StereoDepth.PresetMode.HIGH_ACCURACY,
     )
     stereo.setDepthAlign(dai.CameraBoardSocket.CAM_A)
-    stereo.setOutputSize(*VIDEO_SIZE)
+    if platform == "RVC2":
+        stereo.setOutputSize(*object_detection_nn_archive.getInputSize())
+    # stereo.setLeftRightCheck(True)
+    # stereo.setExtendedDisparity(True)
+    # stereo.setRectification(True)
```
klemen1999, line 69: On these I would say we turn on lr check, extended disp is not needed I think and for rectification is this True by default already?

neural-networks/object-detection/human-machine-safety/utils/__init__.py
klemen1999, line 1: We are now adding nodes that were previously in the host_nodes to each experiment separately. But I feel like this is not the place for them longterm as well. I wonder if we should procede with this or rather leave them in the host_nodes (with symbolic links in each experiment) for now and rather start porting them to dai-nodes instead? @dominik737 for your thoughts on this.

klemen1999 (commented):
Generally LGTM, left some comments and questions

neural-networks/object-detection/human-machine-safety/main.py
kkeroo: Hmm yeah, I can include that.

neural-networks/object-detection/human-machine-safety/utils/__init__.py
kkeroo: I would leave them in the experiment becuase some host nodes needed modifications and didnt work out of the box. Changing them in the host-nodes folder could maybe break some other experiments (it is hard to tell where is the node used). And start porting the nodes to the dai-nodes.

neural-networks/object-detection/human-machine-safety/main.py
```diff
@@ -1,79 +1,128 @@
...
     stereo = pipeline.create(dai.node.StereoDepth).build(
-        left=left_cam.requestOutput(VIDEO_SIZE, fps=FPS),
-        right=right_cam.requestOutput(VIDEO_SIZE, fps=FPS),
+        left=left_cam.requestOutput(
+            object_detection_nn_archive.getInputSize(), fps=FPS_LIMIT
+        ),
+        right=right_cam.requestOutput(
+            object_detection_nn_archive.getInputSize(), fps=FPS_LIMIT
+        ),
         presetMode=dai.node.StereoDepth.PresetMode.HIGH_ACCURACY,
```
dominik737, line 61: Since DAI alpha11 there are new presets and the old two have became deprecated. Now there are options `DEFAULT`, `FACE`, `ROBOTICS`, `HIGH_DETAIL`.

neural-networks/object-detection/human-machine-safety/utils/__init__.py
dominik737: If we have the nodes in the corresponding experiments folder there is advantage that the host node will work and not change, which can also be disadvantage thought (of not getting updates/fixes). Big disadvantage is that when the nodes will be ported to `depthai-nodes` repo and they will have to be unified in a way. This was the way I've done it initially.

Then I received feedback that they should be in one place. That has the advantage of updates/fixes propagating to other experiments. But disadvantage is when you break API you have to be very careful and adjust all the experiments that use the node.

Lastly there was quite a initiative to push all the hostnodes to `depthai-nodes`. Which is the place where they belong. But many of the hostnodes were made quite some time before and with the latest DAI they are useless. Also the porting would take some time. But advantages are big. The nodes can be "pinned" to certain version or semantic versioning (`~=`) can be used to download only non-breaking versions of the package. Only disadvantage is that it takes a bit of time to get new nodes approved and released.

My conclusion would be that, if you can (have time) then try to get all the nodes into `depthai-nodes` repo. Otherwise I think the common `hostnodes` folder is probably better. Because if there are multiple versions in different folders it will be big pain when moving to `depthai-nodes`.

klemen1999: Will be addressed in a separate PR

neural-networks/object-detection/human-machine-safety/main.py
kkeroo: Thanks for info. Fixed in [b450bc3](https://github.com/luxonis/depthai-experiments/pull/592/commits/b450bc346dc5f67c081b15643c18c0244c090d6f).

neural-networks/object-detection/human-machine-safety/main.py
kkeroo: Added left-right check and rectification.

neural-networks/object-detection/human-machine-safety/main.py
kkeroo: There is a bug in the visualizer, for now I left it commented out.

PR #594, opened 2025-01-20
Added gifs to NN readme

klemen1999:
- Added gifs to NN Readme
  - We can roll this our in iterative fashion: When new exepriments roll out we can add their gifs. Goal is for each section to have at least 1 and at most 3 gifs that highlight most impressive examples in the section
- Small updates to some examples so they work with DAIv3a11


klemen1999:
Note: I see that markdown doesn't render as nice on the GH compared to VSCode. I'll work on making it nicer before merge

jkbmrz (approved):
Looking good! Should we add media of the "advanced" examples to the same path (`neural-networks/media`) also? Moreover, should we restrain to using our own footage / royalty free media, and provide links to the sources?

dominik737 (approved):
It looks really nice IMO. Not overwhelming (as the gen2), but still showcases well our models. I would strive to have two GIFs for each category, so the horizontal space usage is maximized.

kkeroo (approved):
Nice!

PR #596, opened 2025-01-22
Gen3: Reidentification Experiments

jkbmrz:
Re-writing human pose / human face re-identification experiments.

Open questions:
- Running on RVC4 over IP, face-re-id runs at 3FPS, and pose-re-id at 4FPS. Is there some way to speed it up? I've tried limiting the amount of info transfer between host and device but cannot get it any lower.
- Should README also contain instructions for standalone mode? Or is including the `.toml` file enough?
- Should GIFs be saved under the common media folder (`neural-networks/media`)?


neural-networks/reidentification/README.md
```diff
@@ -0,0 +1,79 @@
...
+Object re-identification is achieved by calculating cosine similarity between the embeddings.
+
+**WARNING:** The experiment currently works only on RVC4 devices.
+
+## Demo
+
+[![human pose re-identification](media/human_pose_reidentification.gif)](media/human_pose_reidentification.gif)
+[![human face re-identification](media/human_face_reidentification.gif)](media/human_face_reidentification.gif)
+
+[SOURCE](https://www.pexels.com/video/happy-people-walking-on-green-grass-7551577/)
```
klemen1999, line 21: Maybe we make this smaller so it doesn't stand-out as much?
```suggestion
<sup>[Source](https://www.pexels.com/video/happy-people-walking-on-green-grass-7551577/)</sup>
```

neural-networks/reidentification/utils/sync.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
+import depthai as dai
+
+
+class AnnotationSyncNode2(dai.node.ThreadedHostNode):
```
klemen1999, line 5: Why the "2" at the end of the name? I think we can remove it as it feels like there is also 1 somewhere

neural-networks/reidentification/oak_standalone.toml
```diff
@@ -0,0 +1,22 @@
+identifier = "com.luxonis.human-reidentification-example"
+
+entrypoint = ["bash", "-c", "python3 /app/main.py <DET_MODEL> <REC_MODEL> <CSIM> <MEDIA> <FPS_LIMIT> <DEVICE>"]
+
+prepare_container = [
+    { type = "RUN", command = "apt-get update" },
+    { type = "RUN", command = "apt-get install -y python3 python3-pip libglib2.0-0 libgl1-mesa-glx wget git" },
+]
+
+build_steps = [
```
klemen1999, line 10: This should be in requirements.txt and then pip install from requirements (example [here](https://github.com/luxonis/depthai-experiments/blob/feat/update_reidentification_experiments/neural-networks/pose-estimation/human-pose/oak_standalone.toml))

klemen1999 (commented):
Generally LGTM
Let's put this into its own folder (e.g. `human-reidentification`), update the main NN Readme table (links to gen3 example, which models are used and for which platforms it works) and also add these two gifs into the main Readme

neural-networks/reidentification/README.md
jkbmrz: Commited in [2fffd42](https://github.com/luxonis/depthai-experiments/pull/596/commits/2fffd426356c94cc67590f84e3ca1bacc5956d11).

neural-networks/reidentification/oak_standalone.toml
jkbmrz: Added in [8c08194](https://github.com/luxonis/depthai-experiments/pull/596/commits/8c08194203375b01204836622389c1441469a411).

aljazkonec1:
Regarding speedup, I couldnt see any place where the frames would be sent from Host to Device or Device to Host. You can verify this by looking at your network usage, if you send lots of frames, the network should be high. Also how did you connect to the device when measuring FPS? Were you on the same subnet, connected via cable? If you were remote using vpn and/or on wifi then thats the reason for low FPS.

For README, I do add the standalone usage section in my experiments. @klemen1999 if we add for all or just in the general README

I would save gifs in each experiment. GIFs arent shared between different experiments so why would we save them there?

klemen1999:
> For README, I do add the standalone usage section in my experiments. @klemen1999 if we add for all or just in the general README

Correct, standalone usage instructions should be under each experiment, we just don't do `run_standalone.py` anymore as it will be soon deprecated by additonal functionality of oakctl (it was a "quick fix" anyways)

> I would save gifs in each experiment. GIFs arent shared between different experiments so why would we save them there?

+1, gifs under each experiment separately

jkbmrz:
> Also how did you connect to the device when measuring FPS? Were you on the same subnet, connected via cable? If you were remote using vpn and/or on wifi then thats the reason for low FPS.

Yeah, I've connected remote. This must be part of the reason why it's slow!

jkbmrz:
All the requested changes have been added!

klemen1999 (approved):

PR #598, opened 2025-01-22
Unified media for NN experiments

klemen1999:
- All media for exepriments cropped to show only visualizations
- Where external base video/image is shown we add its source (CC: @kkeroo to also add source for pose experiments like we do [here](https://github.com/luxonis/depthai-experiments/tree/gen3/neural-networks/reidentification/human-reidentification#demo))


kkeroo:
Could you add it in the PR?
- [Human-pose](https://www.youtube.com/watch?v=91sd4Jnwgjs)
- [Animal pose](https://www.youtube.com/shorts/LGof_auMHuc)

PR #606, opened 2025-01-28
fix: generic exeriment and re-identification experiment

jkbmrz:
Changes:
- Generic experiment:
    - fix `ImageManipV2.MaxOutputFrameSize()` setting to use stride instead of width - this fixes the experiment for some models that previously errored out with output missmatch error (e.g. `dm-count:shb-426x240` and `rt-super-resolution:50x50`)
    - change the device parameter setting to a more standardized way
    - change host nodes to inherit from `HostNode` instead of `ThreadedHostNode`
- Re-Identification experiment:
    - fix `ImageManipV2.MaxOutputFrameSize()` setting to use stride instead of width
    - error out if running on RVC2 (will be removed once `ImageManipConfigV2` will be able to handle RVC2)
    - remove all `ImageManip` objects (not needed as pipeline is not running on RVC2 currently and `ImageManipV2` will be sufficient after the next DAI update)
    - change `ProcessDetections` host node to inherit from `HostNode` instead of `ThreadedHostNode` (didn't find a way to do it for `AnnotationSyncNode` that syncs outputs of two nn as we need to pull `n` messages from one nn before pulling the next message from the second one)

### DIFF
--- a/neural-networks/generic-example/main.py
+++ b/neural-networks/generic-example/main.py
@@ -51,15 +51,15 @@ with dai.Pipeline(device) as pipeline:
     )
 
     if args.annotation_mode == "segmentation":
-        annotation_node = pipeline.create(SegAnnotationNode)
-        nn_with_parser.passthrough.link(annotation_node.input_frame)
-        nn_with_parser.out.link(annotation_node.input_segmentation)
-        visualizer.addTopic("Video", annotation_node.out, "images")
+        annotation_node = pipeline.create(SegAnnotationNode).build(
+            nn_with_parser.passthrough, nn_with_parser.out
+        )
+        visualizer.addTopic("Video", annotation_node.output, "images")
     elif args.annotation_mode == "segmentation_with_annotation":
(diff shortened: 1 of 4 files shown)


neural-networks/generic-example/main.py
```diff
@@ -8,19 +8,25 @@
...
     model_description.platform = platform
     nn_archive = dai.NNArchive(
         dai.getModelFromZoo(
             model_description,
             apiKey=args.api_key,
         )
     )
+    nn_width = nn_archive.getInputWidth()
+    nn_height = nn_archive.getInputHeight()
+    nn_stride = (nn_width + 7) // 8 * 8  # Align width up to the nearest multiple of 8
```
klemen1999, line 29: This seems a bit hacky. Why is there a +7? Is this unversaly correct?

jkbmrz:
Closing this PR because this is tackled elsewhere (see #615 for the generic experiment updates, and #616 for the re-identification experiment updates).

PR #612, opened 2025-02-05
Gen3: Update DeepSORT Tracking experiment

tadeas0:
Update DeepSORT Tracking experiment to Gen 3.

**Standalone mode is missing, will add it, once oakctl supports command line arguments.**


neural-networks/object-tracking/deepsort-tracking/requirements.txt
```diff
@@ -1,4 +1,6 @@
-depthai~=3.0a1
+--extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-release-local/
+depthai==3.0.0a12
+depthai-nodes==0.1.1
 numpy~=2.0
-scipy
-opencv-python==4.10.0.84
\ No newline at end of file
+opencv-python==4.10.0.84
```
dominik737, line 5: Best to use `opencv-python-headless`, since we don't need the `cv.imshow` anymore.

dominik737 (approved):
Looks good. Works on RVC4 standalone. For peripheral it takes quite a big bandwidth. Did not show me any detections on RVC2 through USB, I also suspect the bandwidth + Visualizer timeout.

tadeas0:
> Looks good. Works on RVC4 standalone. For peripheral it takes quite a big bandwidth. Did not show me any detections on RVC2 through USB, I also suspect the bandwidth + Visualizer timeout.

RVC2 can only run it at ~4 FPS (can be set using `-fps 4` argument) to show detections. Otherwise the embedding OSNet model can't keep up, when there are multiple detections on screen.

neural-networks/object-tracking/deepsort-tracking/main.py
```diff
@@ -1,147 +1,105 @@
...
+        detection_resize.initialConfig.setFrameType(dai.ImgFrame.Type.BGR888i)
+    cam_out.link(detection_resize.inputImage)
 
-    recognition_manip = pipeline.create(dai.node.ImageManip)
-    recognition_manip.initialConfig.setResize(224, 224)
+    detection_nn: ParsingNeuralNetwork = pipeline.create(ParsingNeuralNetwork).build(
+        detection_resize.out, detection_model_archive
+    )
+
+    crop_maker = pipeline.create(DetectionCropMaker).build(
```
klemen1999, line 69: We should use Script node here so there is no device->host->device transfer (example [here](https://github.com/luxonis/depthai-experiments/blob/bed96434f253e4b2bd69990e04a9b58933067f13/neural-networks/pose-estimation/animal-pose/main.py#L83) but you can import that already directly from depthai-nodes [here](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/nodes/utils/detection_config_generator.py), you'll just need to set its main branch in the requirements until we make a new release there)

neural-networks/object-tracking/deepsort-tracking/README.md
klemen1999, line 1: Could you please also update the README in /neural-networks with correct status (emojis) and also add this gif there under object-tracking section (like we did for others)?

neural-networks/object-tracking/deepsort-tracking/utils/detections_recognitions_sync.py
klemen1999, line 1: This syncing seems overenginered. I mean either this is needed and example [here](https://github.com/luxonis/depthai-experiments/blob/bed96434f253e4b2bd69990e04a9b58933067f13/neural-networks/pose-estimation/animal-pose/main.py#L111) ([its host node](https://github.com/luxonis/depthai-experiments/blob/gen3/neural-networks/pose-estimation/animal-pose/utils/annotation_node.py)) is not actually synced or this might be too much and not needed.

dominik737: The `DetectionsRecognitionsSync` node is meant to be a standard node. Which should be ported over to `depthai-nodes`. I think the node as is has many benefits over other approaches.

klemen1999: Why I think this node is not needed is because the main purpose of it is just to sync output from 1st stage NN with output of the 2nd stage NN. But when using HostNode you can just set `self.link_args()` to those and DAI will take care of syncing AFAIK. So we wouldn't even need a separate node, the `DeepsortTracking` node could just take 1st stage NN passthrough and detections and 2nd stage recognition output, set them all to be linked and in its `process()` function operate on them under assumption that DAI has already took care of syncing. Makes sense?

dominik737: I believe there is a distinction between syncing messages 1:1 and syncing 1:n. The former is provided by `self.link_args` the latter is not. Which is where the `DetectionsRecognitionSync` comes in place.

I'm not sure how you could sync 1 `ImgDetections` message with let's say 3 `NNData` objects. The `link_args` will sync the first `ImgDetections` with the first `NNData` object. It does not know of the underlying 3 `ImgDetection` objects in the `ImgDetections`.

tadeas0: I tried to replicate this behavior without the `DetectionsRecognitionSync` node, but could not do it. The issue is, as @dominik737 mentioned, in the fact, that we get multiple recognition messages for a single `ImgDetections` object.

PR #616, opened 2025-02-23
feat: update human-reidentification experiment

jkbmrz:
## Purpose
Adjusting the example to the recent updates in **DAI (3.0.0a13)**:
- `ReplayVideo` node can be used as input to `ParsingNeuralNetwork`;
- `ImageManipV2` works with the `Script` node on RVC2
- host nodes should inherit from `HostNode` instead of `ThreadedHostNode`.

## Specification
Removed `ImageManip` node after the `ReplayVideo` node; removed the option to use the `ImageManip` node with the `Script` node on RVC2; and adjusted the `ProcessDetections` node to inherit from the `HostNode` (yet have to find a way to also do it for the `AnnotationSyncNode` that syncs outputs of two nn as we need to pull n messages from one nn before pulling the next message from the second one. As this is a slightly more complex change, it should probably be tackled in a separate PR for more 2-stage experiments at the same time).

None.

## Deployment Plan
None.

## Testing & Validation
Tested out by running a `luxonis/scrfd-person-detection:25g-640x640` detector and the `luxonis/osnet:imagenet-128x256` recognition model. On RVC2 the code works as expected. On RVC4, no annotations are shown in the visualizer (**TODO: debug why this is the case**). EDIT: On RVC4 the code works as expected as well!


neural-networks/reidentification/human-reidentification/requirements.txt
```diff
@@ -1,3 +1,3 @@
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-release-local/
-depthai>=3.0.0a11
+depthai>=3.0.0a13
 depthai-nodes>=0.1.2
```
klemen1999, line 3: Let's move to depthai-nodes~=0.2.0

klemen1999 (commented):
Generally LGTM but let's make sure the visualizations work as expected as well

neural-networks/reidentification/human-reidentification/requirements.txt
jkbmrz: Added in [0f683b9](https://github.com/luxonis/depthai-experiments/pull/616/commits/0f683b96d7e615f3a88920eeae17c4a36430c426).

jkbmrz:
Visualizations on RVC4 seem to work now. Maybe there were some connection issues when I was trying it out last time.

neural-networks/reidentification/human-reidentification/requirements.txt
klemen1999: Let's use ~=0.2.0 instead >=0.2.0. This way we won't automatically pull e.g version 0.3.0 someday which could have some breaking changes

neural-networks/reidentification/human-reidentification/utils/process.py
```diff
@@ -1,16 +1,16 @@
 import depthai as dai
-from depthai_nodes.ml.messages import ImgDetectionExtended
+from depthai_nodes.message import ImgDetectionExtended, ImgDetectionsExtended
```
klemen1999, line 2: Messages are importable from top level (same as DAI)
```suggestion
from depthai_nodes import ImgDetectionExtended, ImgDetectionsExtended
```

klemen1999 (approved):
LGTM, small nitpick

PR #617, opened 2025-02-24
Fatigue detection update

aljazkonec1:
## Purpose
Update Fatigue detection experiment.

## Specification
Update Fatigue detection to use latest models from HubAI and latest depthai syntax


neural-networks/facial-detection/fatigue-detection/utils/arguments.py
```diff
@@ -0,0 +1,28 @@
+import argparse
+
+def initialize_argparser():
+    """Initialize the argument parser for the script."""
+    parser = argparse.ArgumentParser()
+    parser.description = "OCR example script to show text and show it on a white background. \
```
klemen1999, line 6: Wrong description but we can also remove altogether

klemen1999:
Also please make sure that pre-commit passes before merge (run `pre-commit install` and `pre-commit run --files ...` locally for the first time, after that pre-commit should run before every commit on changed files automatically)

PR #618, opened 2025-02-24
Unifying script node and Removing ImageManip after Replay.

kkeroo:
## Purpose
We are further reducing complexity of the experiments with the new depthai==3.0.0a13 version.

## Specification
We remove explicit ImageManip after the replay node as the Replay node can be sent to the NeuralNetwork node. Also, we unify the content generation for Script node because from now RVC2 also supports ImageManipV2 on device.

No, affected dependencies.

## Deployment Plan
No deployment plan needed.

## Testing & Validation
It was tested on RVC2 and RVC4 with camera and replay node.


klemen1999:
Can we update this PR to dai alpha14?

PR #620, opened 2025-02-27
Gen3: Hand pose

kkeroo:
## Purpose
This PR adds the Hand pose estimation experiment. In the end it also applies some logic to recognize the hand gesture. It can also tell if the hand is left or right. We fix the generic experiment because the init file in utils was missing and detection topic by default was missing.

## Specification
None / not applicable

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested on RVC2 and RVC4 with camera and replay node. Standalone also tested.


klemen1999 (approved):

klemen1999:
Let's also add this experiment to the main README table

PR #623, opened 2025-03-05
DAI nodes v0.2.0 and better E2E testing

klemen1999:
## Purpose

Integration of depthai-nodes after restructure (to be dai-nodes v0.2.0 after we make a release). 
Additionaly we also improve the testing so now all the tests that are meant to pass do so using this branch (action [here](https://github.com/luxonis/depthai-experiments/actions/runs/13676971527)). You can also specify which tests are known to fail on which platform, pythong version or dai version. These then get skipped.

NOTE: A lot of changes have been made to the requirements or type annotations mainly to support python3.8.

TODO:
- Change requirements so we don't pull from depthai-nodes main branch but rather from the official release (once dai-nodes is released)

TODO in the future:
- Explictily add experiments that are known to work on newest DAI versions so they can be used for BOM - either have a separate  "updated_experiments.json" or add those that don't work on newer DAIs to "known_failing_experiments" with specific DAI version specified.  
- Although tests pass this does not 100% mean that the output is expected. We should invest the time to still manually go through experiments and see if the outputs are as expected (mainly on different platforms and DAI versions).

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


dominik737 (approved):
Looks great!

PR #636, opened 2025-03-28
Gen3: Stream manipulation PoE update

tadeas0:
## Purpose
Update stream manipulation PoE experiments to gen3 (`stream-manipulation/poe-tcp-streaming` and `stream-manipulation/poe-mqtt`)

## Testing & Validation
Tested on RVC4.


stream-manipulation/poe-mqtt/utils/paho-mqtt.py
```diff
@@ -1002,14 +1002,15 @@ def json(self):
...
-#     from urllib import request as urllib_dot_request
-# except ImportError:
-#     # Python 2
-#     import urllib as urllib_dot_request
-#     import urlparse as urllib_dot_parse  # type: ignore
+try:
+    # Python 3
+    from urllib import parse as urllib_dot_parse
+    from urllib import request as urllib_dot_request
+except ImportError:
```
klemen1999, line 1009: Does it make sense to keep python2 stuff in anywhere? AFAIK we don't even build DAI for py2 anymore

klemen1999 (commented):
More general comment:
When these new sections are now getting refactored it would be good to regularly update the [tests/expriments_metadata.json](https://github.com/luxonis/depthai-experiments/blob/gen3/tests/experiments_metadata.json), particularly removing known to fail examples from the list when they get updates and are confirmed working. Maybe we also check others that were already updated in the last ~1month (outside of NN folder) and update this file accordingly and regularly otherwise it's harder to keep track of it.
To test it out you can run the action just on a specific exepriment (or folder of exeperiments) by changing the root_dir input parameter

PR #639, opened 2025-04-03
Gen3: Qt gui update

tadeas0:
## Purpose
Replace qt-gui experiment with stereo-runtime-configuration experiment

## Testing & Validation
**Tested on RVC2. Needs to be tested on RVC4.**


dominik737 (commented):
Tested on RVC4 and it works, just had to update the `DepthColorTransform`.

PR #643, opened 2025-04-09
Gen3 depth measurement calc spatial on host

spbui00:


depth-measurement/calc-spatial-on-host/main.py
dominik737, line 1: Please request B&C sensor resolution of `(640, 400)`. The current one throw exception with latest DAI. Should be fixed later, but still.

depth-measurement/calc-spatial-on-host/main.py
```diff
@@ -2,13 +2,15 @@
 
 import cv2
 import depthai as dai
-from calc import HostSpatialsCalc
+from util.roi_control import ROIControl
+from util.arguments import initialize_argparser
 from depthai_nodes.node import ApplyColormap
 from host_node.measure_distance import MeasureDistance
```
dominik737, line 8: Don't use the symbolic link `host_node` folder. Unfortunately current `oakctl` does not support them, thus running this in standalone mode will not work. In the new experiments the host nodes are often in the utils folder with intention to port some of them later.

depth-measurement/calc-spatial-on-host/README.md
```diff
@@ -2,19 +2,63 @@
...
+Moreover, you need to prepare a **Python 3.10** environment with the following packages installed:
+
+- [DepthAI](https://pypi.org/project/depthai/),
+- [DepthAI Nodes](https://pypi.org/project/depthai-nodes/).
+
+You can simply install them by running:
+
+```bash
+pip install -r requirements.txt
+```
```
dominik737, line 27: You don't need this for standalone mode. Move this section into peripheral mode.

depth-measurement/calc-spatial-on-host/requirements.txt
```diff
@@ -1,3 +1,5 @@
-opencv-python
-numpy~=2.0
-depthai==3.0.0a12
\ No newline at end of file
+--extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-release-local/
+depthai==3.0.0a14
```
dominik737, line 2: Use alpha15.

dominik737 (approved):
Works well. Just small notes.

depth-measurement/calc-spatial-on-host/README.md
spbui00: this is taken from the generic example

PR #650, opened 2025-04-17
feat: update human-pose and age-gender experiments

jkbmrz:
## Purpose
Updating the experiments with the new host nodes (`GatherData` and ) helpers (`AnnotationHelper`, `generate_script_content`).

## Specification
human-pose:
- ...

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


jkbmrz:
The `depthai_nodes`' `generate_script_content` function could also be utilized in the following experiments
- pose-estimation/animal-pose,
- object-tracking/deepsort-tracking, 
- 3D-detection/objectron
Can either add it in this or a separate PR.

neural-networks/pose-estimation/human-pose/main.py
```diff
@@ -93,17 +93,27 @@
...
 
+    detections_filter = pipeline.create(ImgDetectionsFilter).build(
+        detection_nn.out, labels_to_keep=valid_labels
+    )
+
+    gather_data_node = pipeline.create(GatherData).build(fps)
+    pose_nn.out.link(gather_data_node.input_data)
+    detections_filter.out.link(gather_data_node.input_reference)
+
+    skeleton_edges = (
```
klemen1999, line 104: I belive this information should be now already part of ImgDetectionExtended message

neural-networks/facial-detection/age-gender/main.py
```diff
@@ -75,21 +91,27 @@
...
+        crop_node.out, rec_model_nn_archive
     )
 
-    sync_node = pipeline.create(DetectionsAgeGenderSync)
-    input_node.link(sync_node.passthrough_input)
-    face_detection_node.out.link(sync_node.detections_input)
-    age_gender_node.getOutput(0).link(sync_node.age_input)
-    age_gender_node.getOutput(1).link(sync_node.gender_input)
+    # gather age-gender info
+    rec_gathered = pipeline.create(GatherData).build(fps, wait_count_fn=return_one)
```
klemen1999, line 100: What is the purpose of this gather? Generally the outputs of the same NN should already be syned between eachother no?

klemen1999 (commented):
> generate_script_content function could also be utilized in some other experiments

Essentially this could be used in every 2stage exepriment no?

neural-networks/pose-estimation/human-pose/main.py
jkbmrz: The detection model here is YOLOv6 so we are dealing with `dai.ImgDetections`. Do you propose we transform to `ImgDetectionsExtended` and merge it together with `Keypoints` into a single message?

neural-networks/facial-detection/age-gender/main.py
jkbmrz: The purpose is to join the two recognitions (age and gender) into a single message that can then be synced with the detections message. Do you maybe see a better way of how this could be done?

neural-networks/pose-estimation/human-pose/main.py
klemen1999: Oh my bad, then this is ok as is I think

neural-networks/facial-detection/age-gender/main.py
klemen1999: Hm I'm thinking how should we set an example for best practice in this case. Maybe creating just GatheredData type manually would be better? Because you only want to "package" up all the things into one message and adding GatherData node just for this seems a bit overkill, adds overhead. But I'm not sure, maybe this is infact the cleanest solution. Thoughts?

jkbmrz: Yeah I agree it is not the cleanest solution. Ideally, `GatherData` would sync `input_data` from multiple streams. But that might be an overkill for a single node.

dominik737: I would advise to not use `GatherData` for this, as it's primary purpose should be to gather data on a non-trivial condition e.g. `NNData` based off `len(ImgDetections.detections)`.

The ways I would go about this are:

1.  Using `dai.node.Sync`, which is made to sync 1:1 messages based on timestamp. The output message is of type `dai.MessageGroup`.
2. Implement a node that can merge inputs into a `dai.MessageGroup` without syncing. The merging could be done based on a lambda as in `GatherData` node.
3. Make a custom type for this particular experiment and return it from a custom node that will merge the messages.
4. Have `outputs` property be part of the `ParsingNeuralNetwork`, where the multiple head outputs are outputted into a single `dai.MessageGroup`.

I personally like the option (4) the most. I think there will hardly be use beside this use-case of merging multiple head outputs into group. Else if such cases are expected I would opt for (2).

kkeroo: I like 4th option. Easy to implement on dai-nodes side and quite useful in such cases (at least 5 multihead models in the ZOO) + can work with `GatherData`.

jkbmrz: I agree. Starting a PR for it [HERE](https://github.com/luxonis/depthai-nodes/pull/208).

PR #656, opened 2025-04-25
feat: abstract input setting

jkbmrz:
## Purpose
There is a lot of code repetition between experiments for input node setting (the procedure of creating either a `dai.node.Camera` or `dai.node.ReplayVideo` node is virtually the same for every experiment).

## Specification
Adding the `create_input_node` function to abstract the input node setting.
TODO:

- utilize the function in all of the experiments within the neural-networks folder,
- discuss if the `create_input_node` function should be moved to `depthai-nodes`.

None

## Deployment Plan
None

## Testing & Validation
Running the experiment(s) using the default settings.

### DIFF
--- a/neural-networks/generic-example/main.py
+++ b/neural-networks/generic-example/main.py
@@ -1,8 +1,8 @@
-from pathlib import Path
 import depthai as dai
 from depthai_nodes.node import ParsingNeuralNetwork, ImgFrameOverlay, ApplyColormap
 
 from utils.arguments import initialize_argparser
+from utils.input import create_input_node
 
 _, args = initialize_argparser()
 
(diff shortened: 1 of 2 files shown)


jkbmrz:
To discuss: Should the `create_input_node` function should be moved to `depthai-nodes`?

tersekmatija:
Hm, I think this is a bit of an early optimization at this stage. I'd avoid from adding those experiments-wide or putting them into depthai-nodes (for now at least). Mainly, I think it makes experiments a bit less clear and straightforward. It is OK for generic-example for now.

But, I agree it's a nice abstraction that has the potential. Maybe something to discuss with DAI team directly if moving it there directly would make sense in the future?

tersekmatija (approved):

jkbmrz:
Okay, I see your point! My motivation for moving it to `depthai-nodes` was to simplify it's import and not so much to migrate it to DAI (might be interesting feature in the future but probably not something of high priority for now). If only using it in the `generic-example` for now, I'd rather keep the location as it is.

klemen1999 (approved):

kkeroo (approved):

PR #657, opened 2025-04-29
feat: update and standardize neural-network experiments

jkbmrz:
## Purpose
Making a general review of the `neural-network` experiments. Aiming to improve the code organization/readability and utilize the most up to date nodes from `depthai-nodes`.

## Specification
       
Modifications:
- Making sure that the experiments following a similar code structure (e.g. first set the NN archives, then inputs, then nodes, etc.) and adding some general code comments;
- Standardizing the setting of FPS (within `ParsingNeuralNetwork.build`, where possible) and media/cam inputs;
- Node utilization:
    - `GatherData`, 
    - `AnnotationHelper`, 
    - `ImgDetectionsFilter`,
    - `ImgDetectionsBridge`

None as all the changes introduced to the experiment arguments are optional and a default value is used if not set.

## Deployment Plan
Wait for a new release of `depthai-nodes` before merging.

## Testing & Validation
All the updated experiments were tested on both `RVC2` and `RVC4` (or the platform the experiment is supported for) using the latest version of `depthai-nodes` (`main` branch, the upcoming `depthai-nodes==0.2.2`) and `depthai` (`v3 develop` branch).


jkbmrz:
`facial-detection/face-mask-detection` -> we currently show wrong labels. Should we add a custom annotation node that will display the correct ones?

jkbmrz:
Points of discussion:

- Some experiments are quite slow (e.g. `facial-detection/emotion-recognition` and `facial-detection/gaze-estimation`). Is there anything we can do to speed them up?
- I’m not getting good results for the `facial-detection/fatigue-detection` experiment (I don’t think I’ve introduced some breaking change because I cannot get good results with the `gen3` version either). Doesn't even work for really simple example as, for example, [HERE](https://www.pexels.com/video/close-up-video-of-a-woman-screaming-8724415/). cc @aljazkonec1 
- `facial-detection/age-gender` and `facial-detection/emotion-recognition` are virtually the same experiment, just different models are used. Should we think about merging them together?
- `pose-estimation/human-pose` has issues with skeleton visualization at higher FPS (the video goes on but the skeletons stop showing). Any ideas what could be the reason? Should we block the input node to not send any new images until the predictions are made?
- `facial-detection/face-mask-detection` currently show wrong labels. Should we add a custom annotation node that will display the correct ones or wait for the upcoming changes in DAI that aim to fix that?
- `counting/cumulative-object-counting`, `counting/depth-people-counting`, and `facial-detection/head-posture-detection` are outdated and don't work out-of-box. I did not touch these experiments in this PR. Do we want to keep and update them, or remove?

kkeroo:
> has issues with skeleton visualization at higher FPS (the video goes on but the skeletons stop showing). Any ideas what could be the reason? Should we block the input node to not send any new images until the predictions are made?

I would say messages get out of sync, since HRNet is not realtime on RVC2 you should set fps limit to lower values (IIRC 10FPS works good).
I think that we should by default set fps on all experiments based on the platform and only override it if user sets specific flag `--fps_limit`. This way first-time user gets the working experiment without figuring out the FPS needed.

jkbmrz:
@kkeroo 

> I think that we should by default set fps on all experiments based on the platform and only override it if user sets specific flag --fps_limit. This way first-time user gets the working experiment without figuring out the FPS needed (see this [example](https://github.com/luxonis/depthai-experiments/blob/3de450f39c4de758b50e52befd8e75940568e19e/neural-networks/object-tracking/deepsort-tracking/main.py#L15))

I agree, makes sense!

> I would add custom node for that so we have fully working example.

Added in [1e730e5](https://github.com/luxonis/depthai-experiments/pull/657/commits/1e730e5d8d7b2cf227b62fa370c0b968110e7aa5) (requires `depthai-nodes` from this [PR](https://github.com/luxonis/depthai-nodes/pull/211))

jkbmrz:
@klemen1999 

> Are they slow on RVC2 or also on RVC4? Because the models themselves should be 80+ FPS on RVC4

I cannot test the exact FPS on RVC4 due to networking lag (but both experiments are pretty slow for me on RVC4). However, using `facial-detection/emotion-recognition` to predict for a single face I get around 5 FPS on RVC2 and I guess that is actually not that bad (considering that the recognition model runs at 8 FPS).

aljazkonec1:
> not getting good results for the facial-detection/fatigue-detection

Can you describe more about the bad results? Does it not even warn for closed eyes / tilted head or does it do it wrong?From my testing, the camera needs to be close to the person and at the same height as the head. The model calculates whether the head is tilted for some time and if the persons eyes are closed for more than ~ one second. The video you linked, the she doesnt move her head down or close her eryes ( apart from in the beginning)

>  And the "head_posture" one was not updated yet (IIRC @aljazkonec1 you've done some work on it already?)
Yes, I've used the same model in gaze estimation so implementing just head posture would be simple

jkbmrz:
@aljazkonec1 

> Can you describe more about the bad results?

I manage to get some tilted head predictions but almost no closed eyes predictions (also for the attached video which seems quite straight-forward to me). Could you maybe run the updated example yourself and evaluate if the performance got worse compared to the last time you tried it out?

PR #664, opened 2025-05-06
Bom integration

danilo-pejovic:
Adding bom integration (echoing ID so workflow can be found from outside and few basic flags)


klemen1999:
Note on failing CI @danilo-pejovic : You should install pre-commit in the environment and run `pre-commit install` so that it is ran automatically on commits

PR #693, opened 2025-06-10
Update animal model.

kkeroo:
## Purpose
We have new animal model in ZOO, so we should replace YOLO with the animal detector.

## Specification
None / not applicable

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested on RVC2, RVC4


klemen1999 (approved):
LGTM but let's also update the example README with correct detection model + main NN readme table with correct model and update the flags showing that it works on RVC4 as well

PR #701, opened 2025-07-03
Gen3 multiple device fusion

spbui00:
# Rewrite to DepthAI v3 + improvements

## FusionManager Node:
- Replaced the simple distance-based grouping with a more resilient time-window-based fusion logic. Detections are now buffered by their timestamp.
- Implemented dynamic timing parameters (`time_window_ms`, `timeout`) that are calculated based on the camera's FPS. This makes the fusion process adaptable to different frame rates and latencies.
- The manager now processes windows of detections that are older than a timeout, ensuring all relevant data from multiple cameras is collected before grouping.
- Added a pruning step (`_prune_redundant_detections`) to ensure each camera contributes at most its highest-confidence detection to a fused group, preventing duplicates.
> Once multiple device sync will be available, this node will be redundant.

## Bird's Eye View Node:
- Focuses on rednering BEV view, it now only receives fused detections (e.g. from FusionManager).


tutorials/multiple-devices/spatial-detection-fusion/utils/annotation_helper.py
klemen1999, line 1: Can we reuse the AnnotationHelper from depthai-nodes?

klemen1999 (commented):
- In requirements we can probably have headless opencv-python instead of the full version?
- Can we update the gif?

dominik737:
@spbui00 I've tried it and it doesn't seem to be working. I get error:
`Exception has occurred: AttributeError       (note: full exception trace is shown but execution is paused at: run)
'depthai.SpatialImgDetection' object has no attribute 'name'
  File "/home/dominik/Dokumenty/source/depthai-experiments/tutorials/multiple-devices/spatial-detection-fusion/utils/fusion.py", line 174, in _transform_detections_to_world
    label=det.name,
          ^^^^^^^^
  File "/home/dominik/Dokumenty/source/depthai-experiments/tutorials/multiple-devices/spatial-detection-fusion/utils/fusion.py", line 80, in _read_inputs
    msg.detections, extrinsics["cam_to_world"], extrinsics["friendly_id"]

            )
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dominik/Dokumenty/source/depthai-experiments/tutorials/multiple-devices/spatial-detection-fusion/utils/fusion.py", line 64, in run (Current frame)
    self._read_inputs()
AttributeError: 'depthai.SpatialImgDetection' object has no attribute 'name'`
What DepthAI version you used to test this? In the `requirements.txt` there is `alpha1`, but it surely is not working on `alpha1` as it doesn't have any visualizer. On `rc2` and `alpha15` it seems to fail on the above exception.

spbui00:
> What DepthAI version you used to test this? In the requirements.txt there is alpha1, but it surely is not working on alpha1 as it doesn't have any visualizer.
Oh i was working on 3.0.0b1, my conda env was incorrect. I'll update it to rc2!

tutorials/multiple-devices/spatial-detection-fusion/README.md
```diff
@@ -1,22 +1,99 @@
...
+> Before you can run this demo you need to calibrate the cameras, the goal is to figure out the exact position and orientation of every camera relative to each other, so the application can fuse their individual views into a single, unified "world". Go to [multi-cam-calibration](../multi-cam-calibration) and generate a calibration file for each camera. Once calibration is complete, do not move your cameras. If you do, you will need to re-calibrate. Make sure that the `calibration_data_dir` in the [`config.py`](config.py) is set correctly.
 
-| key | action            |
-| :-- | :---------------- |
-| `q` | quit              |
-| `d` | toggle depth view |
+The system employs a *main device* architecture. Upon launch, the first available OAK camera is designated as the central processing hub.
 
-## Usage
+- **Main Device**: This camera is responsible for more than just its own detections. It also runs the FusionManager and BirdsEyeView HostNodes. This means all the complex data merging and visualization logic happens on this one device.
```
dominik737: I'm thinking a bit if this is true. When you have a `dai.node.HostNode` or `dai.node.ThreadedHostNode` in peripheral mode this code technically runs on host (which is your PC).

tutorials/multiple-devices/spatial-detection-fusion/requirements.txt
```diff
@@ -1,4 +1,6 @@
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-release-local/
-depthai~=3.0.0a1
-opencv-python==4.10.0.84
-numpy>=1.22
\ No newline at end of file
+depthai==3.0.0rc2
+depthai-nodes==0.3.1
```
dominik737: I would use `~=` for `depthai-nodes` specifically. Unlike in `depthai` there is not history of breaking changes being release to a minor version. Also I think it's convention among other experiments as well.

tutorials/multiple-devices/spatial-detection-fusion/utils/config.py
```diff
@@ -0,0 +1,17 @@
...
+
+nn_model_slug = "luxonis/yolov6-nano:r2-coco-512x288"
+
+nn_input_size = (512, 288)  # Should match the chosen YOLOv6 model's input size
+
+# specific labels to show in the BEV / empty labels to show all
+bev_labels = []  # Show all labels by default
+# bev_labels = ["bottle"]
+
+distance_threshold_m = 0.5  # Distance threshold for grouping detections in meters
```
dominik737: In Python it's common convention to write all constants with `ALL_UPPERCASE` syntax. This then get's picked up by static analyzers as well.

tutorials/multiple-devices/spatial-detection-fusion/utils/fusion.py
```diff
@@ -0,0 +1,238 @@
...
+        )  # time window for grouping near-simultaneous detections
+        self.timeout = frame_time_ms / 1000  # timeout for fusion in seconds
+        self.latest_device_timestamp_ms = 0
+
+        self.distance_threshold_m = (
+            distance_threshold  # Distance threshold for grouping detections in meters
+        )
+
+    def run(self):
+        while True:
```
dominik737: You can replace with while `self.isRunning()`.

tutorials/multiple-devices/spatial-detection-fusion/utils/fusion.py
```diff
@@ -0,0 +1,238 @@
...
+            self._read_inputs()
+            self._process_buffer()
+            time.sleep(0.002)
```
dominik737: I believe this is causing the annotations to being sent at 500FPS. For me the browser is lagging quite a lot, because it is unable to render it at such speeds. You can select bigger number. Preferable even tied to FPS (`1/FPS`, `0.75/FPS`).

tutorials/multiple-devices/spatial-detection-fusion/utils/fusion.py
```diff
@@ -0,0 +1,238 @@
...
+                return
+
+            groups = self._group_detections(all_detections_in_window)
+            pruned_groups = self._prune_redundant_detections(groups)
+
+            data_bytes = pickle.dumps(pruned_groups)
+            buffer = dai.Buffer()
+            buffer.setData(bytearray(data_bytes))  # type: ignore
+            buffer.setTimestamp(datetime.timedelta(milliseconds=start_ts))
+            self.output.send(buffer)
```
dominik737: I think you don't need to pickle and unpickle the class. You should be able to send the class as message directly. Just inherit `dai.Buffer`.

dominik737 (commented):
Nice, it runs on RVC4. The issue I had was likely related to my device and was resolved after OS update. Please just take a look at the FPS at which the annotations are rendering for me it caused quite a lag on FE.

tutorials/multiple-devices/spatial-detection-fusion/main.py
```diff
@@ -1,388 +1,185 @@
...
-            filtered_devices.append(d)
+    left_cam = pipeline.create(dai.node.Camera).build(
+        dai.CameraBoardSocket.CAM_B,
+    )
+    right_cam = pipeline.create(dai.node.Camera).build(
+        dai.CameraBoardSocket.CAM_C,
+    )
+    stereo = pipeline.create(dai.node.StereoDepth).build(
+        left=left_cam.requestOutput(nn_input_size, fps=fps_limit),
+        right=right_cam.requestOutput(nn_input_size),
```
dominik737: Here the fps limit is not set.

tutorials/multiple-devices/spatial-detection-fusion/README.md
spbui00: You're right .The `dai.node.HostNode` and `dai.node.ThreadedHostNode` do execute their code on the host, not the OAK in the peripherel mode. The concept of a "main device" is still relevant though. The `HostNode` have to be registered to a specific device's pipeline. So in this case, the main device's pipeline is simply the one that acts as a central communication hub rather than processing hub as I stated. I'll modify the readme to be more accurate. Thanks!

PR #710, opened 2025-07-15
Increasing FPS on RVC2 due to parser optimizations.

kkeroo:
## Purpose
We optimized some parsers so we can increase FPS on some examples.

## Specification
Increased FPS on age-gender, head-posture-detection, and hand-pose. Not all face-detection examples are possible to increase becasue in some the bottleneck is 2nd stage model for example in emotion recognition.

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable


klemen1999 (commented):
Let's change the requirements to new depthai-nodes version

PR #720, opened 2025-07-21
Gen3 depth measurement box measurement

spbui00:


streaming/poe-mqtt/oakapp.toml
```diff
@@ -1,13 +1,37 @@
+<<<<<<< HEAD:depth-measurement/3d-measurement/box-measurement/oakapp.toml
+identifier   = "com.luxonis.box-measurement"
+app_version  = "0.1.0"
+=======
 identifier = "com.luxonis.poe-mqtt"
 app_version = "1.0.0"
+>>>>>>> main:streaming/poe-mqtt/oakapp.toml
```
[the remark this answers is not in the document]
dominik737: @spbui00 please resolve. Unresolved git conflict was commited somehow.

depth-measurement/3d-measurement/pointcloud/device-pointcloud/main.py
```diff
@@ -0,0 +1,97 @@
...
+def create_xyz(width, height, camera_matrix):
+    xs = np.linspace(0, width - 1, width, dtype=np.float32)
+    ys = np.linspace(0, height - 1, height, dtype=np.float32)
+
+    # Generate grid by stacking coordinates
+    base_grid = np.stack(np.meshgrid(xs, ys))  # WxHx2
+    points_2d = base_grid.transpose(1, 2, 0)  # 1xHxWx2
+
+    # Unpack coordinates
+    u_coord: np.array = points_2d[..., 0]
```
dominik737: The `np.array` is only used to allocate the array, but the underlying type is `NDArray` from `numpy.typing`.

depth-measurement/3d-measurement/pointcloud/device-pointcloud/requirements.txt
dominik737, line 1: Please resolve the conflicts

depth-measurement/3d-measurement/box-measurement/utils/box_processing_node.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def _annotate_detection(
+        self, det: ImgDetectionExtended, idx: int, mask: np.ndarray, pcl, pcl_colors
+    ):
+        """Draw all annotations (mask, 3D box fit, bounding box + label) for a single detection."""
+        self._draw_mask(mask, idx)
+
+        self._fit_cuboid(idx, mask, pcl, pcl_colors)
+
+        self._draw_box_and_label(det)
```
dominik737: Remove the spaces.

depth-measurement/3d-measurement/box-measurement/main.py
```diff
@@ -1,87 +1,110 @@
...
+_, args = initialize_argparser()
+
+NN_WIDTH, NN_HEIGHT = 512, 320
+INPUT_SHAPE = (NN_WIDTH, NN_HEIGHT)
+
+IMG_WIDTH, IMG_HEIGHT = 640, 400
+CAMERA_RESOLUTION = (IMG_WIDTH, IMG_HEIGHT)
+
+visualizer = dai.RemoteConnection(httpPort=8082)
+device = dai.Device(dai.DeviceInfo(args.device)) if args.device else dai.Device()
```
dominik737: You can create the `dai.Device` from ip just with `dai.Device(ip)`.

depth-measurement/3d-measurement/box-measurement/main.py
```diff
@@ -1,87 +1,110 @@
...
+device.setIrLaserDotProjectorIntensity(1.0)
+device.setIrFloodLightIntensity(1)
```
dominik737: I'm not sure whether the flood light helps with the precision. Also use `1` or `1.0` for both so it's unified.

depth-measurement/3d-measurement/box-measurement/oakapp.toml
```diff
@@ -0,0 +1,28 @@
...
+"""
+bash -c '
+DIR=$(python3 -c "import open3d, pathlib, depthai; print(pathlib.Path(open3d.__file__).parent)") &&
+echo ">> linking into $DIR" &&
+for lib in libEGL.so.1 libGL.so.1 libGLESv2.so.2; do
+  ln -sf /usr/lib/aarch64-linux-gnu/$lib $DIR/$lib &&
+  echo "Linked $lib"
+done
+'
+"""
```
dominik737: What is this script doing?

depth-measurement/3d-measurement/box-measurement/oakapp.toml
```diff
@@ -0,0 +1,28 @@
...
+]
+
+entrypoint = ["bash", "-c", "python3 -u /app/main.py"]
+
+environment = {PYTHONUNBUFFERED="1", OPEN3D_CPU_RENDERING = "true", PIP_BREAK_SYSTEM_PACKAGES = "1"}
```
dominik737: What is the purpose of these variables?

depth-measurement/3d-measurement/box-measurement/main.py
```diff
@@ -1,87 +1,110 @@
...
-        min_box_size=args.min_box_size,
+    if platform == dai.Platform.RVC2:
+        nn.setNNArchive(nn_archive, numShaves=7)
+
+    nn._parsers[0].setConfidenceThreshold(0.7)
+    nn._parsers[0].setIouThreshold(0.5)
+    nn._parsers[0].setMaskConfidence(0.5)
+
+    box_processing = BoxProcessingNode()
+    box_processing.intrinsics = read_intrinsics(device, NN_WIDTH, NN_HEIGHT)
```
dominik737: Why is the node not added to the pipeline?

depth-measurement/3d-measurement/box-measurement/main.py
```diff
@@ -1,87 +1,110 @@
...
+    right = p.create(dai.node.Camera).build(dai.CameraBoardSocket.CAM_C)
+
+    stereo = p.create(dai.node.StereoDepth).build(
+        left=left.requestOutput(CAMERA_RESOLUTION, fps=args.fps_limit),
+        right=right.requestOutput(CAMERA_RESOLUTION, fps=args.fps_limit),
+    )
+    stereo.initialConfig.setMedianFilter(dai.StereoDepthConfig.MedianFilter.KERNEL_7x7)
+    stereo.setDefaultProfilePreset(dai.node.StereoDepth.PresetMode.DEFAULT)
+    stereo.enableDistortionCorrection(True)
+    stereo.setExtendedDisparity(True)
```
dominik737: The kernel filter is not supported for the `DEFAULT` profile likely. The extended disparity has too many bits which results in
> [1944301041DE4D1300] [3.1] [4.717] [StereoDepth(3)] [error] Maximum disparity value '1520' exceeds the maximum supported '1024' by median filter. Disabling median filter!
[1944301041DE4D1300] [3.1] [4.796] [StereoDepth(3)] [error] Maximum disparity value '1520' exceeds the maximum supported '1024' by median filter. Disabling median filter!

dominik737: For RVC4 the combination seems to be fine, no warning is produced.

depth-measurement/3d-measurement/box-measurement/oakapp.toml
spbui00: This is a fix for an open3d linking issue, mostly on ARM devices. It finds the open3d install directory and then symlinks the system's graphics libs (EGL, GL) into it so open3d can render correctly.

depth-measurement/3d-measurement/box-measurement/oakapp.toml
spbui00: `PYTHONUNBUFFERED`: Makes sure Python logs appear immediately. By default, Python holds onto output and prints it in chunks for efficiency. In a long-running application or a Docker container, this means you might not see your logs or print statements for a long time.

`OPEN3D_CPU_RENDERING`: this just forces open3d to use the cpu for rendering

`PIP_BREAK_SYSTEM_PACKAGES`: its a flag to override the default pip behavior that prevents modifying system
  packages. Since we're in a container, it's safe to do this.

All those flags are from the oakapp.toml of the original project, if we have not had any problem with python logging, or pip install, then only then `OPEN3D_CPU_RENDERING` is necessary imo

dominik737: The `python3 -u` already does what the `PYTHONUNBUFFERED="1"` is doing. The `--break-system-packages` is used above so I guess the `PIP_BREAK_SYSTEM_PACKAGES` is also redundant.

depth-measurement/3d-measurement/pointcloud/device-pointcloud/requirements.txt
```diff
@@ -0,0 +1,9 @@
+<<<<<<< HEAD:depth-measurement/3d-measurement/pointcloud/device-pointcloud/requirements.txt
+--extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-release-local/
+depthai~=3.0.0a1
+=======
+depthai==3.0.0rc2
+>>>>>>> main:depth-measurement/3d-measurement/rgbd-pointcloud/requirements.txt
```
dominik737: Here is also confict as well.

fraymackers:
> > Hello, I tested the code with my Oak-D Pro W (RCV2), I have a first error caused by extended disparity `Maximum disparity value '1520' exceeds the maximum supported '1024' by median filter. Disabling median filter!`. Then, the camera freeze and I have no more images.
> > And when I disable the extended disparity, the error disappeared but I have always the freeze of the camera after some frames without error.
> 
> Hello @fraymackers, I believe there are two separate issues you are experiencing. First is that the kernel filter is not supported for the set disparity value. This should not prevent you from running the example though. This means there is likely another issue in your case. I was testing the code on `OAK-D Pro` and I'm not experiencing the issue. Perhaps it could be the `W` variant. Could you also share which DepthAI version were you using and how do you power the device? CC: @moratom

Hi, thank you for your reply. Ipower my camera with Y-plug and I set IR led to 0 to be sure. Here are version I'm using: 
image

fraymackers:
> Hi @fraymackers . I have tested the experiment on my Oak-D Pro W (RCV2) and it worked fine. (The new pushes include disabling the Median Filter for RVC2 so the errors messages about the disparity values should be gone).
> 
> I suggest you create an **Issue**. The underlying culprit could be related to your device specifically.

Hi @spbui00 , thank you for your test, I'll open an issue.

streaming/poe-mqtt/oakapp.toml
```diff
@@ -3,7 +3,7 @@ app_version = "1.0.0"
 
 prepare_container = [
     { type = "RUN", command = "apt-get update" },
-    { type = "RUN", command = "apt-get install -y python3-pip" },
+    { type = "RUN", command = "apt-get install -y --no-install-recommends python3-pip libegl1-mesa libgomp1 libgl1-mesa-glx libglib2.0-0 libusb-1.0-0 usbutils" },
```
klemen1999: Was there some error that resulted in the need for these extra packages? Because nothing was caught with the runnability tests.

dominik737: @spbui00 Isn't this this some issue that was brought in by the merging issue you told me about?

depth-measurement/3d-measurement/box-measurement/main.py
spbui00: I missed that one when rewriting, good catch ;)

streaming/poe-mqtt/oakapp.toml
spbui00: These are the conflicts happened when I merged main this branch, which was outdated. I resolved these with a new rebase

depth-measurement/3d-measurement/box-measurement/README.md
```diff
@@ -4,67 +4,58 @@ This example showcases one possible approach for measuring the size of a box usi
 
 ![demo](https://github.com/luxonis/oak-examples/blob/master/gen2-multiple-devices/box-measurement/img/demo.gif)
 
-_For a more accurate version of multi-cam box measurement check out the [demo here](https://github.com/luxonis/oak-examples/tree/master/gen2-multiple-devices/box-measurement)._
+> **⚠️ Known issue:** Due to current CuboidFitter algorithm limitations, measurements may occasionally jump or skip between frames. Active stereo is crucial for stable results, and you may see some instability, especially on RVC2 devices.
```
dominik737: The text is good now. I'm just thinking that it might be good to mention that the instability can be caused by not enough power.

depth-measurement/3d-measurement/box-measurement/main.py
```diff
@@ -1,87 +1,113 @@
...
+NN_WIDTH, NN_HEIGHT = 512, 320
+INPUT_SHAPE = (NN_WIDTH, NN_HEIGHT)
+
+IMG_WIDTH, IMG_HEIGHT = 640, 400
+CAMERA_RESOLUTION = (IMG_WIDTH, IMG_HEIGHT)
+
+visualizer = dai.RemoteConnection(httpPort=8082)
+device = dai.Device(args.device) if args.device else dai.Device()
+device.setIrLaserDotProjectorIntensity(1.0)
+device.setIrFloodLightIntensity(1.0)
```
dominik737: Please create a thread whether the floodlight actually helps. My understanding is that it rather makes the dot projector patter less visible (they both work with IR).

depth-measurement/3d-measurement/box-measurement/main.py
```diff
@@ -1,87 +1,113 @@
...
+    color_output.link(manip.inputImage)
+
+    nn = p.create(ParsingNeuralNetwork).build(nn_source=nn_archive, input=manip.out)
+
+    if platform == dai.Platform.RVC2:
+        nn.setNNArchive(nn_archive, numShaves=7)
+
+    nn._parsers[0].setConfidenceThreshold(0.7)
+    nn._parsers[0].setIouThreshold(0.5)
+    nn._parsers[0].setMaskConfidence(0.5)
```
dominik737: Please use `nn.getParser()`. It is designed such that the default index is 0.

depth-measurement/3d-measurement/box-measurement/oakapp.toml
dominik737: Can you remove those two as they are redundant?

depth-measurement/3d-measurement/box-measurement/requirements.txt
```diff
@@ -1,5 +1,5 @@
-depthai==3.0.0rc2
-opencv-python==4.10.0.84
+depthai==3.0.0rc3
+depthai-nodes==0.3.1
 numpy>=1.22
-open3d~=0.18
-matplotlib
+open3d==0.19.0
```
dominik737: Let's put back the `~=0.18` please. It seems that the 0.19 is not built for arm (doesn't work for standalone). But for Python 3.12 it's nice.

depth-measurement/3d-measurement/box-measurement/main.py
dominik737: Generally speaking refrain from using protected fields (starting with `_`) outside the class that defines them.

spbui00:
> Small nitpick that I saw with the whole `depth-measurement` folder: The app identifiers in these examples don't follow the pattern exactly. It should be `com.example.<folder name>.<example name>`. So in this case it would be: `com.example.depth-measurement.box-measurement` Could you update this example and all others in the `depth-measurement` folder accordingly @spbui00 ?

box measurement and rgbd pointcloud are in subfolder 3d-measurement/, so should the identifier be `com.example.depth-measurement.3d-measurement.[example-name]`?

PR #721, opened 2025-07-22
Add an example C++ application

themarpe:
## Purpose
To expand oak-examples with native app development (C++) as well

@asdwjure please finalize this and make 2 examples, one UVC and one "generic" (perhaps mimicing some other simpler Python example). CC: @klemen1999

### DIFF
--- a/tutorials/depthai-cpp-demo/oakapp.toml
+++ b/tutorials/depthai-cpp-demo/oakapp.toml
@@ -1,0 +1,42 @@
+# Application metadata
+identifier = "com.example.streaming.uvc"
+app_version = "1.0.0"
+
+# Command to run when the container starts
+entrypoint = ["bash", "-c", "/app/uvc-start.sh start"]
+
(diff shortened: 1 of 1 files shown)


tersekmatija (changes requested):
Let's make sure we somehow clearly differentiate between `py` and `cpp` examples as well, otherwise this can quickly become confusing. Potentially using `cpp` prefix for all C++ examples could work. Alternatively, they are placed in a separate folder?

We should document this in README as well.

klemen1999:
IMO we place all CPP examples in a different, top level folder. We can simply call it `cpp` or `cpp-examples`? And if there is more of them we can also split them again into different subfolders inside that one.

This way they can be also easily removed from the python test suite - longerterm it would be nice to also have some cpp runability tests for these ones though if someone wants to add them.

moratom:
Should be "replaced" by https://github.com/luxonis/oak-examples/pull/798 we should still add an UVC example as well though.

PR #723, opened 2025-07-25
Add new tof-pointcloud example

lnotspotl:
## Purpose
This PR adds an example showcasing the workings of our newly updated ToF node.


depth-measurement/3d-measurement/tof-pointcloud/README.md
```diff
@@ -0,0 +1,28 @@
+# ToF pointcloud demo
+
+This example shows how to use `depthai`'s ToF node and visualize some of its outputs, including `depth` and `depthRaw`.
+Moreover, this demo also turns the `depth` images into pointclouds which are then visualized, side-by-side with the `depth` frames, using Open3D.
+Finally, through an interactive GUI, one can adjust the ToF's underlying filters, allowing one to get intution of the filtering techniques used or to tune the settings for one's specific needs.
+
+**NOTE**: This example requires a ToF camera. You can get one from the official [Luxonis store](https://shop.luxonis.com/products/oak-d-sr-poe).
+
+## Demo
+
```
klemen1999, line 8: Missing (when added let's make sure if it is a GIF it is max ~10MB, the lower the better)

depth-measurement/3d-measurement/tof-pointcloud/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Usage
```
klemen1999, line 11: Can we follow more unified README template? Example [here](https://github.com/luxonis/oak-examples/tree/main/neural-networks/generic-example) ( should be actually same on any other exampe currently in `main`):
- Usage section you can mostly copy paste but edit the arguments section (copy from python3 main.py --help)
- This is RVC2 only example right? In that case having `Peripheral Mode` and then `Installation` and `Examples` sections under it makes sense

depth-measurement/3d-measurement/tof-pointcloud/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Installation
+
+The demo was tested with **Python 3.12** and several external dependencies installed.
```
klemen1999: We test examples with pytron 3.8, 3.10 and 3.12. And we "promote" 3.10 as the "main one" (check Installation sections in other examples). 
Is there a particular reason why this wouldn't work on 3.8 or 3.10?

depth-measurement/3d-measurement/tof-pointcloud/README.md
dominik737: I think it would be nice if we could get a gif here as is in the other experiments.

depth-measurement/3d-measurement/tof-pointcloud/requirements.txt
```diff
@@ -0,0 +1,6 @@
+--extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
+depthai==3.0.0rc3.dev0+ed76d4ac4dab4fd6f3e897f8d4fa5d606497dc32
+
+numpy==2.2.6
+open3d==0.19.0
+opencv-python==4.12.0.88
```
klemen1999: Can we use these dependencies so they are compatible with e.g. py3.8 as well?
```
numpy>=1.22
open3d~=0.18
opencv-python~=4.10.0
```

depth-measurement/3d-measurement/tof-pointcloud/README.md
dominik737, line 11: We can more or less copy the section from the other experiments, just remove the standalone section and mention that it is not working in standalone due to lack of RVC4 ToF.

depth-measurement/3d-measurement/tof-pointcloud/requirements.txt
```diff
@@ -0,0 +1,6 @@
+--extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
+depthai==3.0.0rc3.dev0+ed76d4ac4dab4fd6f3e897f8d4fa5d606497dc32
```
dominik737: I'm just curios here, is the ToF node available only in this snapshot and later?

depth-measurement/3d-measurement/tof-pointcloud/tof.py
klemen1999, line 1: Can we rename to `main.py`? This is what gets automatically then grabbed by the [runnability tests](https://github.com/luxonis/oak-examples/actions/workflows/tests.yml)

klemen1999, line 1: General comment:
Is there a particular reason we don't use visualizer instead of `tkinter` and `cv2.imshow`? Are there maybe limitations of what can be displayed there and what you want to display in this example?
CC: @dominik737 interested if you think this could be done with visualizer or not really atm

dominik737, line 1: @klemen1999 Do we still need to name the experiment files `main.py` to get picked up by tests?

klemen1999 (commented):
Some general comments, didn't run the example yet (not sure if I'll be able to do so soon since I don't have access to the SR camera)

depth-measurement/3d-measurement/tof-pointcloud/requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+
+numpy==2.2.6
```
dominik737: Let's unify with other experiment numpy requirements. The advantage of the syntax we have is it supports older python versions. Beside that we can just find&replace if we want to make bulk changes.

depth-measurement/3d-measurement/tof-pointcloud/requirements.txt
dominik737: Same for the opencv

depth-measurement/3d-measurement/tof-pointcloud/requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+open3d==0.19.0
```
dominik737: I've had issues with running the `0.19` on arm. In my experience it is best to do `~=0.18`.

depth-measurement/3d-measurement/tof-pointcloud/tof.py
dominik737, line 1: Would it be possible to use Visualizer instead of the `cv2` and `tkinter`? We have an example where there are also [runtime controls](https://github.com/luxonis/oak-examples/tree/main/depth-measurement/stereo-runtime-configuration).

dominik737: Yea, that was my comment too actually. I would think it is doable. We support pointcloud rending two ways:

- via RGBD node output
- sending depth stream and encoded color stream separately

dominik737: When I think about it, even better if the ToF node output message would be added to protobuf serializations and would be easily connectable to visualizer. Similar to RGBD node.

dominik737: I think both could be used in this case. The RGBD node might be even better as it's simple to understand (drawback being bigger bandwidth usage).

klemen1999: Yes, we search for presence of `main.py` and `requirements.txt` in a folder and if these are there then we test it. Its not a bulletproof solution but I don't have a better idea. And it also forces the oak-examples to have some uniformity.

depth-measurement/3d-measurement/tof-pointcloud/README.md
lnotspotl: Left if out deliberately, @DavidZah will create a nice short GIF and upload it here :)

depth-measurement/3d-measurement/tof-pointcloud/requirements.txt
lnotspotl: Yes, the ToF node got an overhaul, now featuring on-host running filters for improved accuracy and precision. The version above is the first one that had the improved ToF node included. Now that that feature has been merged to `main`, I'll change the required version to `depthai==3.0.0rc4` which is already on pypi: https://pypi.org/project/depthai/3.0.0rc4/

depth-measurement/3d-measurement/tof-pointcloud/README.md
lnotspotl: I added a tiny gif

depth-measurement/3d-measurement/tof-pointcloud/requirements.txt
lnotspotl: The version of depthai specified in requirements.txt is `depthai~=3.0`

depth-measurement/3d-measurement/tof-pointcloud/README.md
lnotspotl: This example indeed should work on all these python version. In fact, any python version in the range of 3.8 - 3.13 should be fine.

depth-measurement/3d-measurement/tof-pointcloud/tof.py
lnotspotl: I am not sure it would be a quick change and there is a blog post to be released soon. Maybe we keep it as is right now and do a quick swap (I can take care of that myself, would love to know about the visualizer more) after that post?

klemen1999: Ok sounds good, we can change it to use visualizer instead in a separate PR

PR #737, opened 2025-08-05
Fixed padding for yolo-world postprocessing

klemen1999:
## Purpose

Fixed edge case where padding was not applied correctly in the new `tokenizers` code.
Tested with previously failing example:
`class_names = ["a", "airplane", "refrigerator", "dog", "microwave oven", "cup"]`

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/custom-frontend/dynamic-yolo-world/backend/src/utils/helper_functions.py
+++ b/custom-frontend/dynamic-yolo-world/backend/src/utils/helper_functions.py
@@ -36,10 +36,13 @@ def extract_text_embeddings(class_names, max_num_classes=80):
         save_path="tokenizer.json",
     )
     tokenizer = Tokenizer.from_file(tokenizer_json_path)
+    tokenizer.enable_padding(
+        pad_id=tokenizer.token_to_id("<|endoftext|>"), pad_token="<|endoftext|>"
+    )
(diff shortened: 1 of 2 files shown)


custom-frontend/dynamic-yolo-world/backend/src/utils/helper_functions.py
```diff
@@ -36,10 +36,13 @@ def extract_text_embeddings(class_names, max_num_classes=80):
         save_path="tokenizer.json",
     )
     tokenizer = Tokenizer.from_file(tokenizer_json_path)
+    tokenizer.enable_padding(
+        pad_id=tokenizer.token_to_id("<|endoftext|>"), pad_token="<|endoftext|>"
```
sokovninn, line 40: LGTM. Just one question: why is a specific `pad_token` used here instead of the default `[PAD]`?

klemen1999: From my tests of printing out what padding is used in `transformers` I saw that 49407 id is used which is the `<Iendoftext|>` token. So I kept the same behaviour.

PR #740, opened 2025-08-08
Add YOLOE to custom frontend demo

sokovninn:
## Purpose

This change integrates the YOLOE model into the backend, providing enhanced object detection capabilities with efficient visualization features like instance segmentation.

## Specification

* Backend now supports loading and running the YOLOE model.
* Documentation updated to reflect the addition of YOLOE as an option in the dynamic YOLO-world example.
* Disables image uploading for YOLOE as it is not supported in the current setup.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


dominik737 (approved):
I've tried it in standalone and it works well. If we would want to polish it more we could disable the upload image section if the model is yoloe.

PR #753, opened 2025-08-27
Feat/yaml way

klemen1999:
## Purpose

- Added yaml way of loading the models to all examples
- Removed the useCached reference - this is True by default

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


aljazkonec1 (approved):

klemen1999:
For tracking: Tests passed but when a very small fix was pushed the whole tests were cancelled since it only affected one example and that was tested separately

PR #761, opened 2025-09-10
Enable visual prompts for YOLOE in the custom frontend demo

sokovninn:
## Purpose

- Add visual-and-text-prompt YOLOE variant and precision control.
- Enable image-mask prompting path (YOLOE visual encoder) alongside existing CLIP path.
- Add high-res visualization and video encoding
- Add bbox drawing functionality
- Add YOLO-World FP16 version
- Add notification toasts

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/custom-frontend/dynamic-yolo-world/README.md
+++ b/custom-frontend/dynamic-yolo-world/README.md
@@ -22,7 +22,7 @@ Here is a list of all available parameters:
 					FPS limit. (default: None)
 -ip IP, --ip IP       IP address to serve the frontend on. (default: None)
 -p PORT, --port PORT  Port to serve the frontend on. (default: None)
--n MODEL_NAME, --model-name MODEL_NAME
+-m MODEL, --model MODEL
 					Name of the model to use: yolo-world, yoloe, or yoloe-image (default: yolo-world)
 --precision PRECISION
 					Model precision for YOLOE models: int8 (faster) or fp16 (more accurate) (default: int8)
--- a/custom-frontend/dynamic-yolo-world/README.md
+++ b/custom-frontend/dynamic-yolo-world/README.md
@@ -62,7 +62,11 @@ oakctl app run .
 
 Once the app is built and running you can access the DepthAI Viewer locally by opening `https://<OAK4_IP>:9000/` in your browser (the exact URL will be shown in the terminal output).
 
-This will run the example with default argument values (YOLO-World model). If you want to change these values you need to edit the `oakapp.toml` file (refer [here](https://docs.luxonis.com/software-v3/oak-apps/configuration/) for more information about this configuration file).
+This will run the example with default argument values (YOLO-World model). If you want to change these values you need to edit the `backend-run.sh` file to pass the arguments to the backend. Example:
+
+```bash
+python3.12 /app/backend/src/main.py --model yoloe --precision fp16 --fps-limit 10 --device <DEVICE_ID>
+```
 
 ### Remote access
 
--- a/custom-frontend/dynamic-yolo-world/README.md
+++ b/custom-frontend/dynamic-yolo-world/README.md
@@ -1,6 +1,6 @@
 # Dynamic YOLO World/YOLOE
 
-This example demonstrates an advanced use of a custom frontend. On the DepthAI backend, it runs either the **YOLO-World** (default) or **YOLOE** model on-device, with configurable class labels and confidence threshold — both controllable via the frontend.
+This example demonstrates an advanced use of a custom frontend. On the DepthAI backend, it runs either the **YOLO-World** (default), **YOLOE**, or **YOLOE-Image** model on-device, with configurable class labels and confidence threshold — both controllable via the frontend.
 The frontend, built using the `@luxonis/depthai-viewer-common` package, displays a real-time video stream with detections. It is combined with the [default oakapp docker image](https://hub.docker.com/r/luxonis/oakapp-base), which enables remote access via WebRTC.
 
 > **Note:** This example works only on RVC4 in standalone mode.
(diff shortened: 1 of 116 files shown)


custom-frontend/dynamic-yolo-world/README.md
```diff
@@ -23,15 +23,22 @@ Here is a list of all available parameters:
 -ip IP, --ip IP       IP address to serve the frontend on. (default: None)
 -p PORT, --port PORT  Port to serve the frontend on. (default: None)
 -n MODEL_NAME, --model-name MODEL_NAME
-					Name of the model to use: yolo-world or yoloe (default: yolo-world)
+					Name of the model to use: yolo-world, yoloe, or yoloe-image (default: yolo-world)
+--precision PRECISION
+					Model precision for YOLOE models: int8 (default) or fp16. fp16 disables input quantization.
```
dominik737: I think it would be great if there is brief description what are the impacts of choosing the int8 and fp16.

custom-frontend/dynamic-yolo-world/README.md
```diff
@@ -23,15 +23,22 @@ Here is a list of all available parameters:
...
 
 ### Model Options
 
-This example supports two different YOLO models:
+This example supports three different YOLO models:
 
-- **YOLO-World** (default): An open-vocabulary object detection model that supports both text-based class definitions and image-based prompting (upload an image to detect similar objects)
-- **YOLOE**: A fast and efficient object detection model with enhanced visualization features including instance segmentation
+- **YOLO-World** (default): Open-vocabulary detection with text prompts and optional image prompting (CLIP visual encoder).
+- **YOLOE**: Fast detection with enhanced visualization, including instance segmentation.
```
dominik737: I think it should be mentioned here explicitly that the input image box does not work for this model.

dominik737 (approved):

custom-frontend/dynamic-yolo-world/README.md
```diff
@@ -23,15 +23,22 @@ Here is a list of all available parameters:
 -ip IP, --ip IP       IP address to serve the frontend on. (default: None)
 -p PORT, --port PORT  Port to serve the frontend on. (default: None)
 -n MODEL_NAME, --model-name MODEL_NAME
-					Name of the model to use: yolo-world or yoloe (default: yolo-world)
+					Name of the model to use: yolo-world, yoloe, or yoloe-image (default: yolo-world)
+--precision PRECISION
```
klemen1999: If we are more confident in FP16 model and it runs with ~30FPS then I'm fine using FP16 as default as well

custom-frontend/dynamic-yolo-world/README.md
```diff
@@ -23,15 +23,22 @@ Here is a list of all available parameters:
 -ip IP, --ip IP       IP address to serve the frontend on. (default: None)
 -p PORT, --port PORT  Port to serve the frontend on. (default: None)
 -n MODEL_NAME, --model-name MODEL_NAME
-					Name of the model to use: yolo-world or yoloe (default: yolo-world)
+					Name of the model to use: yolo-world, yoloe, or yoloe-image (default: yolo-world)
```
klemen1999: Considering testing so far maybe we set YOLO-E as the default?

custom-frontend/dynamic-yolo-world/README.md
klemen1999, line 1: Given this is a standalone only example and CLI parameters need to be (at least currently) manually specified in the backend-run.sh file I think we can add extra instructions regard this part so its more clear how one can switch which model/mode they want to use

custom-frontend/dynamic-yolo-world/backend/src/main.py
```diff
@@ -33,9 +33,25 @@
     raise ValueError("This example is supported only on RVC4 platform")
 
 frame_type = dai.ImgFrame.Type.BGR888i
-text_features = extract_text_embeddings(
-    class_names=CLASS_NAMES, max_num_classes=MAX_NUM_CLASSES, model_name=args.model_name
-)
+# choose initial features: text for yolo-world/yoloe, visual for yoloe-image
+if args.model_name == "yoloe-image":
+    import numpy as np
```
klemen1999: I'd move this to the top of the file even though it's only use conditionally but we already have it in requirements.txt anway

custom-frontend/dynamic-yolo-world/backend/src/utils/arguments.py
```diff
@@ -51,11 +51,19 @@ def initialize_argparser():
     parser.add_argument(
         "-n",
         "--model_name",
-        help="Name of the model to use: yolo-world or yoloe",
+        help="Name of the model to use: yolo-world, yoloe or yoloe-image",
```
klemen1999: I'd use `-m`, `--model` for this to be more consistent with other examples

custom-frontend/dynamic-yolo-world/backend/src/main.py
```diff
@@ -116,6 +138,11 @@
 
     def class_update_service(new_classes: list[str]):
         """Changes classes to detect based on the user input"""
+        if args.model_name == "yoloe-image":
+            print(
+                "Class update is disabled in yoloe-image mode. Upload a new image prompt instead."
```
klemen1999: In case this happens can we give some feedback to the user in frontend? Like a toast error mesage? I'm not sure if such connection from service -> frontend is possible? CC: @dominik737 if you maybe konw?

custom-frontend/dynamic-yolo-world/README.md
```diff
@@ -62,7 +62,11 @@ oakctl app run .
 
 Once the app is built and running you can access the DepthAI Viewer locally by opening `https://<OAK4_IP>:9000/` in your browser (the exact URL will be shown in the terminal output).
 
-This will run the example with default argument values (YOLO-World model). If you want to change these values you need to edit the `oakapp.toml` file (refer [here](https://docs.luxonis.com/software-v3/oak-apps/configuration/) for more information about this configuration file).
+This will run the example with default argument values (YOLO-World model). If you want to change these values you need to edit the `backend-run.sh` file to pass the arguments to the backend. Example:
+
+```bash
+python3.12 /app/backend/src/main.py --model yoloe --precision fp16 --fps-limit 10 --device <DEVICE_ID>
```
klemen1999: I belive --device shouldn't be needed. The example only runs in standalone so device is inherently the device that the example is running on. I think we can remove the device flag all together for this example (even from main.py arguments)

custom-frontend/dynamic-yolo-world/README.md
sokovninn: @klemen1999 The only factor that is stopping me from doing that is that there are separate YOLOE models for text and image input, so to switch between them, the app needs to be restarted.

klemen1999: I see. But if YOLOE text performance is magnitude better from YoloWorld then I'd use that one for the default. But definetly we should add also on frontend some note on how one can switch between them easily and what each of these modes (models) is good for

custom-frontend/dynamic-yolo-world/backend/src/main.py
sokovninn: @klemen1999 @dominik737 sending data from frontend to backend only works for me, not the other way around.

klemen1999: Yeah I think we can leave as is for now and improve later on if needed (and is supported - because here I don't think this connection back is yet supported, just wanted to check)

PR #771, opened 2025-10-27
Scoped tbs

danilo-pejovic:
additing additional options flag for BOM MODE so we can test selected subset of devices

Commented out strict mode for now (As I dont think it is used, let me know if better workaround for 10 param limit is needed)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,11 +6,8 @@ on:
       distinct_id:
         description: 'run identifier'
         required: false
-      depthai_version:
-        description: 'Version of DepthAI to install for the test run. If specified the version will override version specified in requirements.txt'
-        required: false
-      depthai_nodes_version:
-        description: 'Version of depthai-nodes to install for the test run. If specified the version will override version specified in requirements.txt'
+      dai_versions:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -37,14 +37,17 @@ on:
...
         default: "."
         required: false
-      strict_mode:
-        description: "If set to 'yes', tests will fail on DepthAI warnings."
-        type: choice
-        options:
-          - "yes"
-          - "no"
-        default: "no"
+      # strict_mode:
```
klemen1999: So this needs to be commented out because there is a max number of params workflow can have?

danilo-pejovic: Yea - github doesnt allow more than 10. I can couple dai version and nodes version into 1 param (make it json input) going around this limitation if needed.

klemen1999: Yeah this might be better if possible because for bigger updates I do use strict testing to also check no new warnings are getting printed (potential silent errors)

danilo-pejovic: Ok, changed - if tests pass (both ones started on this branch and ones incoming from bom) it should be good to merge

PR #772, opened 2025-11-07
multi-cam stitching and nn detections example

jz592058:
Adding a new example which stitches camera stream from multiple cameras and runs yolo model over the resulting panorama. It uses tiling to make it possible for the panorama to fit inside NN input resolution.

### DIFF
--- a/tests/constants.py
+++ b/tests/constants.py
@@ -20,7 +20,7 @@ KNOWN_FAILING = {
         "reason": "Not ported to latest DAI yet",
         "platform": "all",
     },
-    "tutorials/multiple-devices/multiple-device-stich-nn": {
+    "tutorials/multiple-devices/multiple-device-stitch-nn": {
         "reason": "Test suite doesn't support multi-device testing",
         "platform": "all",
     },


tutorials/multiple-devices/multiple-device-stitch-nn/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Here is a list of all available parameters:
+
+```
+-fps FPS_LIMIT, --fps_limit FPS_LIMIT
+                    FPS limit for the model runtime. (default: 20)
+-is INPUT_SIZE, --input_size INPUT_SIZE
+                    Input video stream resolution. {2160p, 1080p, 720p, 480p, 360p} (default: 360p)
+```
+
```
klemen1999, line 33: Nitpick: Lets put this whole bottom section in `## Peripheral Mode` heading. So:
```
## Peripheral Mode
#### Installation
....
### Examples
....
```

tutorials/multiple-devices/multiple-device-stitch-nn/main.py
```diff
@@ -0,0 +1,135 @@
...
+        outputs.append(output)
+
+    script_path = pathlib.Path(
+        __file__
+    ).parent.resolve()  # get absolute path of this script
+    platform = device.getPlatform()  # get platform
+    # Get model based on platform type from yaml file
+    model_description = dai.NNModelDescription.fromYamlFile(
+        f"{script_path}/depthai_models/yolov6_nano.{platform.name}.yaml"
+    )
```
klemen1999: Is this script path needed? Afaik this should work like this since .fromYamlFile() looks in `depthai_models` dir by default? Not 100% sure though.
```suggestion
    platform = device.getPlatform()  # get platform
    # Get model based on platform type from yaml file
    model_description = dai.NNModelDescription.fromYamlFile(
        f"yolov6_nano.{platform.name}.yaml"
    )
```

tutorials/multiple-devices/multiple-device-stitch-nn/utils/stitch.py
klemen1999, line 1: Nitpick: I'd go a bit lighter on the comments and remove some of the obvious ones e.g. "# get the frame from input", "# assert that it has correct type", etc. Same note for other files as well

tutorials/multiple-devices/multiple-device-stitch-nn/main.py
jz592058: yes its needed if you want to be able to run this script from anywhere. It checks inside `depthai_models` but only relative to the run path. So for example: qr with tiling example which does not have a full, absolute path implemented, has a problem:
```
(venv) mk@mkw ~/Documents/Git/oak-examples (nn-stitching) $ python3 tutorials/qr-with-tiling/main.py 
To connect to the DepthAI visualizer, open http://localhost:8082 in your browser
In case of a different client, replace 'localhost' with the correct hostname
Creating pipeline...
Traceback (most recent call last):
  File "/home/mk/Documents/Git/oak-examples/tutorials/qr-with-tiling/main.py", line 20, in <module>
    model_description = dai.NNModelDescription.fromYamlFile(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Model file not found: `depthai_models/qrdet_nano.RVC2.yaml` | If you meant to use a relative path, prefix with ./ (e.g. `./qrdet_nano.RVC2.yaml`) | Also, make sure the file exists. Read the documentation for more information.
(venv) mk@mkw ~/Documents/Git/oak-examples (nn-stitching) $ cd tutorials/qr-with-tiling/
(venv) mk@mkw ~/Documents/Git/oak-examples/tutorials/qr-with-tiling (nn-stitching) $ python3 main.py 
To connect to the DepthAI visualizer, open http://localhost:8082 in your browser
In case of a different client, replace 'localhost' with the correct hostname
Creating pipeline...
Pipeline created.
```

klemen1999: Hm true but I would argue that all instructions (e.g. all commands in readme like `python3 main.py` or `oakctl app run .`) are already assuming that your current working directory is the one of the specific example. So because of that I would go forward with this assumption also in the scripts. This is what is done in other examples as well.

jz592058: @klemen1999  hmm, ok, maybe there is something in the background that I'm not seeing, but just to push the point a bit here: what if  the `.fromYamlFile()` were changed to look at path of the script instead of the path from where the user is running it? Then you wouldn't need to assume the current working directory but could run from anywhere as well as keep all the examples as they are?

klemen1999: That would be nice but this is then a DepthAI V3 feature request.

tests/constants.py
```diff
@@ -20,6 +20,10 @@
         "reason": "Not ported to latest DAI yet",
         "platform": "all",
     },
+    "tutorials/multiple-devices/multiple-device-stich-nn": {
+        "reason": "Stitching can't be performed on cameras since multiple streams need to be joined together",
+        "mode": ["standalone"],
```
klemen1999:
```suggestion
        "platform": ["rvc4"],
```
Because multiple RVC4 devices are not automatically available through HIL testing

tutorials/multiple-devices/multiple-device-stitch-nn/README.md
```diff
@@ -0,0 +1,68 @@
...
+## Peripheral Mode
+
+### Installation
```
aljazkonec1, line 36: I would move installaition before Usage section as first you install then use. I would also remove the `Peripheral mode` title as the example only runs in peripheral, no need to have it as a separate section.

tutorials/multiple-devices/multiple-device-stitch-nn/README.md
```diff
@@ -0,0 +1,68 @@
...
+python3 main.py -fps 10
+```
+
+This will run the example at 10 FPS.
+
+```bash
+python3 main.py -is 720p
+```
+
+This will run the example with resolution 720p. Stitching, tiling and YOLO detections can be costly on the processing resources - especially with larger number of cameras connected - and the output FPS will depend on the CPU power. If output FPS is too low, try lowering the resolution.
```
aljazkonec1: I would specify this part in the "Limitations" part and specify that the stitching is done on the host computer which is the main bottle neck if you have a large numnber of devices connected.

tutorials/multiple-devices/multiple-device-stitch-nn/README.md
klemen1999: The structure is IMO good as is already because it follows what other examples have. The reason for "Usage" before "Installation" is that installation is mode dependant (peripheral vs standalone) where as usage is the same for both. 
I agree it does look a bit weirder in this case where only peripheral is supported but it's worth "price to pay" for having it more unified across examples.

tests/constants.py
```diff
@@ -20,6 +20,10 @@
         "reason": "Not ported to latest DAI yet",
         "platform": "all",
     },
+    "tutorials/multiple-devices/multiple-device-stich-nn": {
```
klemen1999:
```suggestion
    "tutorials/multiple-devices/multiple-device-stitch-nn": {
```

PR #777, opened 2025-11-18
Hotfix the GIF of stitching-nn example

jz592058:
Previous gif ended in darkness instead of looping. The new gif is also smaller.


klemen1999:
Skipping tests as it is just a GIF change

PR #778, opened 2025-11-21
Compatibility with depthai-nodes==0.3.5

kkeroo:
## Purpose
We introduced breaking changes in the dai-nodes==0.3.5 and we need to fix oak-examples.

## Specification
Breaking changes were in `script_content_generator` and `Tiling` node. We fixed that by:
- removing deprecated field in `script_content_generator`, and adding `ImgDetectionFilter` in two examples, to filter out valid labels. In `human-pose` we already have `ImgDetectionFilter`.
- Modifying `qr-code-with-tiling` example because `Tiling` node was completely rewritten. It needs dai==3.1.0, because in 3.0.0 there was some error with `Script` node.

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/tests/constants.py
+++ b/tests/constants.py
@@ -80,6 +80,10 @@ KNOWN_FAILING = {
         "mode": "all",
         "platform": "all",
     },
+    "tutorials/qr-with-tiling": {
+        "reason": "Missing bindings for RVC2 in Script node.",
+        "platform": ["rvc2"],
+    },
 }
 
 IGNORED_WARNINGS = [


tutorials/qr-with-tiling/main.py
klemen1999, line 1: Because of the node changes this example is no longer supported on RVC2. We should mark this in the README, the table in `/tutorials` and excludi it from the RVC2 tests (chage in `constants.py`)

klemen1999 (commented):
LGTM with small note

> Should we upgrade all oak-examples to dai==3.1.0 and dai-nodes==0.3.5?

No, we'll wait for DAI 3.2 to do this since that one will actually have some breaking changes

tutorials/qr-with-tiling/main.py
kkeroo: Thanks for noticing, updated readmes in [6dad0d6](https://github.com/luxonis/oak-examples/pull/778/commits/6dad0d6ca93051b1da56465f9326268c6f2fdb46).

PR #779, opened 2025-11-24
Add basic ros examples

Serafadam:
## Purpose

ROS app examples
## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/apps/ros/ros-driver-basic/README.md
+++ b/apps/ros/ros-driver-basic/README.md
@@ -1,6 +1,7 @@
 # ROS driver basic example
 
 ![Image](./media/ros_main.png)
+
 This is an example of ROS driver being run as an OAK 4 app. It launches ROS driver in standalone and publishes RGB, Stereo and IMU data.
 These topics can be accessible for viewing and/or further processing directly on host.
 It is based on ROS2 Kilted, but you should be able to subscribe to topics in other distributions such as Humble or Jazzy.
--- a/apps/ros/ros-driver-custom-workspace/README.md
+++ b/apps/ros/ros-driver-custom-workspace/README.md
@@ -1,6 +1,7 @@
 # ROS driver custom workspace example
 
 ![Image](./media/ros_main.png)
+
 This is an example of an app that allows creating custom workspaces that contain your nodes which can be serve as entrypoint to your whole ROS2 project.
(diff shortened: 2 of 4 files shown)


apps/ros/ros_driver_basic_example/README.md
```diff
@@ -0,0 +1,70 @@
...
+- ROS2 (Humble, Jazzy, or Kilted)
+- Rviz2
+
+## Setup Instructions
+
+1. Install the required ROS packages:
+
+   ```bash
+   sudo apt update
+   sudo apt install ros-<distro>-image-transport-plugins ros-<distro>-rviz2
```
jz592058:
```suggestion
   sudo apt install ros-$ROS_DISTRO-image-transport-plugins ros-$ROS_DISTRO-rviz2
```

and i would change all other occourances of `<distro>` to `$ROS_DISTRO` as well

apps/ros/ros_driver_basic_example/oakapp.toml
```diff
@@ -0,0 +1,26 @@
+identifier = "ros.driver.example"
```
klemen1999: Nitpick: 
- Let-s make directorie names with `-` instead of `_` to match the style of other examples (`ros_driver_basic_exampple` -> `ros-driver-basic-example`)
- For oak-examples we have identifiers based on the example location. In this case it would be:
`com.example.apps.ros.ros-driver-basic-example`

klemen1999: Same comment for other examples as well

apps/ros/ros_driver_basic_example/oakapp.toml
```diff
@@ -1,5 +1,5 @@
-identifier = "ros.driver.example"
-app_version = "1.0.0"
+identifier = "com.luxonis.ros.driver.example"
```
klemen1999: We don't want to have `com.luxonis` actually pushed on oak-examples `main` because this identified is reserved for official Luxonis Team. So if one would try to push examle as is to Hub it would fail until they would change the identifier - which is not what we want, examples should be usable as is

PR #791, opened 2025-12-08
Dino poc

FSAFTik:
## Purpose
Added Dino Tracker application

## Specification

Added Dino Tracking Application to apps directory

Doesn't depend on any unreleased technologies

## Deployment Plan

Has to be merged to main and deployed in the Hub. The only problematic behavior is "freezing" in synchronization when some nodes work slower then expected due to big amount of detection (bug on DepthAI, reported)

## Testing & Validation

Was manually tested for a long time with different inputs

### DIFF
--- a/apps/dino-tracking/media/DinoGraph.png
+++ b/apps/dino-tracking/media/DinoGraph.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:030721787d357dcbf95a117a09b6cf79d82b9574a394a95354630d56ef6b985f
-size 1425246
+oid sha256:1061827932afbab898cd4570ab383ebdf2f3012ae64c6e3a59f31e79a122c1fe
+size 39312


apps/dino-tracking/backend/src/core/neural_network_builder.py
```diff
@@ -0,0 +1,47 @@
...
+
+        w, h = self.archive.getInputSize()
+        self.input_width = int(w)
+        self.input_height = int(h)
+
+    @property
+    def input_size(self) -> tuple[int, int]:
+        return self.input_width, self.input_height
+
+    def build(self, video_full: dai.Buffer) -> dai.Node.Output:
```
PetrNovota: `video_full` is not a `Buffer` it is a `dai.Node.Output`

apps/dino-tracking/backend/src/core/neural_network_builder.py
```diff
@@ -0,0 +1,47 @@
...
+        manip = self.pipeline.create(dai.node.ImageManip)
```
PetrNovota: the manip here is not optimal. The preferred way is to use `camera.requestOutput`to request different resolutions you need. Typically, one for the live stream and one for a nn input.

This way is better because it runs faster then using manip

apps/dino-tracking/backend/src/core/neural_network_builder.py
```diff
@@ -0,0 +1,47 @@
+import depthai as dai
+from depthai_nodes.node import ParsingNeuralNetwork
+
+
+class NNBuilder:
```
PetrNovota: this entire class can be replaced with
```
pipeline.create(ParsingNeuralNetwork).build(camera_out, nn_model_name)
```

apps/dino-tracking/backend/src/core/dino_similarity/prompting/click_processor.py
```diff
@@ -0,0 +1,70 @@
+import cv2
+import numpy as np
+import depthai as dai
+from .selection_processor import SelectionProcessor
+from .frame_cache import FrameCache
+from .click_state import ClickState
+
+
+class ClickProcessor:
```
PetrNovota: This should not be named `ClickProcessor` This is a `ObjectSelection` handler.

the main method `process_pending_click` is doing 3 things.
1. processes a potential click
2. fetches data from cache
3. changes the selected area

additionally, the `update_cache_from_msgs` updates the cache, but this should be internal and hidden, because the correct behaviour of this class now depends on the correct order of calling `process_pending_clicks` and `update_cache_from_msgs`

The question is, which is the behavior the user of this class is really interested in. The user is interested in the **object selection** - that is the primary high level behaviour this class implements.

So the correct public method would be something like this
```
def update(self, frame: dai.ImgFrame, segmentation: dai...)
    click = process_click()
    if not click:
        return
    segmentation_full_resolution = self._cache_full_res.get()
    segmentation_fast_sam_resolution = self._cache_fast_sam_res.get()
    self._select_object(
        click=click,
        segmentation_full_resolution= segmentation_full_resolution,
        segmentation_fast_sam_resolution= segmentation_fast_sam_resolution,
    self._cache_full_res.update(frame, segmentation)
    self. _cache_fast_sam_res.update(frame, segmentation)
)
```

apps/dino-tracking/backend/src/core/dino_similarity/dino_selection_node.py
```diff
@@ -0,0 +1,81 @@
+import cv2
+import numpy as np
+import depthai as dai
+
+from depthai_nodes.node import BaseHostNode
+from .prompting.click_processor import ClickProcessor
+from .similarity_engine.dino_similarity_engine import DinoSimilarityEngine
+
+
+class DinoSelectionNode(BaseHostNode):
```
PetrNovota: This node produces a heat map. so it should be named based on its behaviour as seen from the outside of this class. `HeatMapNode`could work

apps/dino-tracking/backend/src/core/detections_tracking/heatmap_detection_node.py
```diff
@@ -0,0 +1,97 @@
+import cv2
+import numpy as np
+import depthai as dai
+from depthai_nodes.node.base_host_node import BaseHostNode
+
+
+class HeatmapDetectionNode(BaseHostNode):
```
PetrNovota: a better name is `HeatmapToBoundingBox`

apps/dino-tracking/backend/src/core/base_service.py
```diff
@@ -0,0 +1,18 @@
...
+
+PayloadT = TypeVar("PayloadT", bound=BaseModel)
+
+
+class BaseService(ABC, Generic[PayloadT]):
+    """
+    Base class for all backend services with payload validation.
+    """
+
+    NAME: str
```
PetrNovota: This is not a good base class, it doesn't achieve much. it doesn't enforce the `NAME` and `PAYLOAD_MODEL`, each subclass has to implement the same validation logic.

One possible implementation with example usage would be:

Base class

```
from __future__ import annotations

from abc import ABC, abstractmethod
from typing import Any, ClassVar, Generic, Optional, TypeVar

from pydantic import BaseModel, ValidationError

PayloadT = TypeVar("PayloadT", bound=BaseModel)


class BaseService(ABC, Generic[PayloadT]):
    """
    Service boundary:
      - called with JSON-ish payload (typically dict)
      - validates using PAYLOAD_MODEL
      - calls handle_typed() with a Pydantic model (or None)
    """

    NAME: ClassVar[str]
    PAYLOAD_MODEL: ClassVar[type[PayloadT]]

    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)

        # enforce required class attrs (fail fast at import time)
        if cls is BaseService:
            return

        if not isinstance(getattr(cls, "NAME", None), str) or not cls.NAME.strip():
            raise TypeError(f"{cls.__name__} must define NAME: str (non-empty).")

        if not hasattr(cls, "PAYLOAD_MODEL"):
            raise TypeError(
                f"{cls.__name__} must define PAYLOAD_MODEL: type[BaseModel]."
            )

        m = cls.PAYLOAD_MODEL
        if m is not None and (not isinstance(m, type) or not issubclass(m, BaseModel)):
            raise TypeError(
                f"{cls.__name__}.PAYLOAD_MODEL must be a pydantic BaseModel subclass or None."
            )

    # This is what you register: callback(json) -> json
    def __call__(self, payload: Any) -> dict:
        try:
            model = self._validate(payload)
            return self.handle_typed(model)
        except ValidationError as e:
            # consistent error schema for all services
            self.on_validation_error(e)
            return {"ok": False, "error": e.errors()}
        except Exception as e:
            # consistent internal error schema (optional)
            self.on_internal_error(e)
            return {"ok": False, "error": str(e)}

    def _validate(self, payload: Any) -> Optional[PayloadT]:
        if self.PAYLOAD_MODEL is None:
            return None
        return self.PAYLOAD_MODEL.model_validate(payload)

    # hooks (optional) for logging
    def on_validation_error(self, e: ValidationError) -> None:
        pass

    def on_internal_error(self, e: Exception) -> None:
        pass

    @abstractmethod
    def handle_typed(self, payload: Optional[PayloadT]) -> dict:
        """Implement service logic using validated payload."""
        raise NotImplementedError
```

Usage:
```
from pydantic import BaseModel, Field

class ThresholdUpdatePayload(BaseModel):
    threshold: float = Field(..., ge=0.0, le=1.0)


class ThresholdService(BaseService[ThresholdUpdatePayload]):
    NAME = "Threshold Update Service"
    PAYLOAD_MODEL = ThresholdUpdatePayload

    def __init__(self, heatmap_det):
        self._heatmap_det = heatmap_det

    def on_validation_error(self, e):
        self._heatmap_det._logger.info(f"Validation error in ThresholdService: {e}")

    def handle_typed(self, payload: ThresholdUpdatePayload | None) -> dict:
        assert payload is not None
        self._heatmap_det._logger.info(
            f"Updating confidence threshold to {payload.threshold}"
        )
        self._heatmap_det.set_confidence_threshold(payload.threshold)
        return {"ok": True, "confidence": self._heatmap_det.get_confidence_threshold()}
```

usage in code
```
service = ThresholdService(heatmap_det)
visualizer.registerService(service.NAME, service)  # <-- register the callable object
```

apps/dino-tracking/backend/src/core/dino_similarity/dino_selection_node.py
```diff
@@ -0,0 +1,81 @@
...
+          * run DINO tracker (two refs: init + track)
+          * cos = combine_alpha * cos_track + (1 - combine_alpha) * cos_init
+      - output heatmap (0..255)
+    """
+
+    def __init__(self):
+        super().__init__()
+
+        self.clicks: ClickProcessor = ClickProcessor()
+        self.tracker: DinoSimilarityEngine = DinoSimilarityEngine()
```
PetrNovota: this cannot be named `tracker`. I think it is not a tracker and in the `main.py` there is also a tracker created
```
        tracker_factory = Tracker(
            pipeline=pipeline,
            detections=heat_map_to_bbox.out,
            frame=rgb_full_res,
        )

        tracker = tracker_factory.build()
```

the tracker in the `main.py` is something else than this tracker and it is likely the correct tracker

apps/dino-tracking/backend/src/core/neural_network_builder.py
FSAFTik: removed the whole class, it was redundant

apps/dino-tracking/backend/src/core/neural_network_builder.py
FSAFTik: replaced

apps/dino-tracking/backend/src/core/neural_network_builder.py
FSAFTik: removed the class

apps/dino-tracking/backend/src/core/dino_similarity/prompting/click_processor.py
FSAFTik: Moved logic to SelectionMaskNode and simplified, this class doesnt exist anymore

apps/dino-tracking/backend/src/core/detections_tracking/heatmap_detection_node.py
FSAFTik: But the node does indeed produce DetectionsObjects, maybe HeatmapToDetectionsNode is a better name here?

apps/dino-tracking/backend/src/core/dino_similarity/dino_selection_node.py
FSAFTik: Node refactored into several nodes, the one that produces heatmap is now SimilarityHeatmapNode

apps/dino-tracking/backend/src/core/base_service.py
FSAFTik: Thanks for a reference, did change base service according to feedback

apps/dino-tracking/backend/src/core/dino_similarity/dino_selection_node.py
FSAFTik: refactored the whole node, no tracker then dai.node.ObjectTracker from main exists

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+from core.annotations.outlines_overlay_node import OutlinesOverlayNode
+from core.object_selection.FE_prompt_services.click_prompt_service import (
+    ClickPromptService,
+)
+from core.annotations.dino_annotation_node import DinoAnnotationNode
+from core.encoder import Encoder
+from core.state_service import StateService
+from core.dino_similarity.similarity_heatmap_node import SimilarityHeatmapNode
+from core.object_selection.selection_mask_node import SelectionMaskNode
+
```
PetrNovota, line 36: unsorted imports. Please sort.
Can `pip install isort` and `isort .` in `src` folder to do automatically

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+        dino_rgb = camera.requestOutput(
+            size=constants.nn.dino.input_size,
+            type=dai.ImgFrame.Type.BGR888i,
+            fps=constants.camera.fps,
+        )
+
+        fastsam_nn = pipeline.create(ParsingNeuralNetwork).build(
+            input=fastsam_rgb, nn_source=constants.nn.segmentation.model_name
+        )
+        segmentation_out = fastsam_nn.out
```
PetrNovota: inconsistent naming. decide whether you want to call it `fastsam` or `segmentation`

PetrNovota: dont create an alias for `fastsam_nn.out`. its understandable on its own

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+            input=dino_rgb,
+            nnArchive=dai.NNArchive(
+                dai.getModelFromZoo(
+                    dai.NNModelDescription(
+                        model=constants.nn.dino.model_name, platform=platform
+                    )
+                )
+            ),
+        )
+        dino_out = dino_nn.out
```
PetrNovota: this is not needed, correct name would be `dino_nn_out` and that is the same as `dino_nn.out` and is therefore not needed and wanted

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+
+        selection_node = pipeline.create(SelectionMaskNode).build(
```
PetrNovota: `MaskSelectionNode` is a more natural way, it reads easier

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+            frame_in=rgb_sensor,
+            segmentations=segmentation_out,
+        )
+
+        prompt_service = ClickPromptService(selection_node)
+        clear_service = ClearSelectionService(selection_node)
```
PetrNovota: Inconsistent naming. It has to be either:
`ClickPromtService` and `ClearSeletectionPromtService`

or
`ClickService` and `ClearSeletectionService`

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+            ),
+        )
+        dino_out = dino_nn.out
+
+        selection_node = pipeline.create(SelectionMaskNode).build(
+            frame_in=rgb_sensor,
+            segmentations=segmentation_out,
+        )
+
+        prompt_service = ClickPromptService(selection_node)
```
PetrNovota: Name base on what this is really doing. I would use
`ObjectSelection`

best imo would be to use

`ObjectSelectionPrompt`
`ClearSelectionPrompt`

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+        dino_grid = pipeline.create(DinoGridExtracorNode).build(dino_in=dino_out)
+
+        reference_node = pipeline.create(ReferenceFromSelectionNode).build(
+            mask_in=selection_node.out,
+            dino_in=dino_grid.out,
+            dino_input_size=constants.nn.dino.input_size,
+        )
+
+        adaptive_reference_node = pipeline.create(
+            AdaptiveReferenceVectorNode, constants.reference_adaptation
```
PetrNovota: inconsistent naming. Decide if its `reference_vector` or just `reference`. Imo better is `reference_vector` 
-->
`reference_vector_node`
`ReferenceVectorFromSelectionNode`
`adaptive_reference_vectore_node`
`AdaptiveReferenceVectorNode`

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+        reference_node.out.link(adaptive_reference_node.init_input)
+
+        similarity_node = pipeline.create(SimilarityHeatmapNode).build(
+            references_in=adaptive_reference_node.out,
+            grid_in=dino_grid.out,
+            frame_in=rgb_sensor,
+        )
+        similarity_node.vector_out.link(adaptive_reference_node.feedback_input)
+
+        heatmap_det = pipeline.create(HeatmapToDetectionsNode).build(
```
PetrNovota: no shortcuts! It has to be
`heatmap_to_detection_node`
I would personally omit the `node` in all the names, but if you are using it, you have to be consistent

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
...
+        tracker = tracker.build()
+
+        outlines_node = pipeline.create(OutlinesOverlayNode).build(
+            frame=rgb_sensor,
+            segmentation=segmentation_out,
+        )
+        outlines_service = OutlinesTriggerService(outlines_node)
+        visualizer.registerService(outlines_service.NAME, outlines_service)
+
+        annot_node = pipeline.create(DinoAnnotationNode).build(
```
PetrNovota: I am losing it here, DON'T shorten the variable names!!!!

apps/dino-tracking/backend/src/main.py
```diff
@@ -0,0 +1,174 @@
+from core.dino_similarity.reference_vectors.adaptive_reference_vector_node import (
+    AdaptiveReferenceVectorNode,
+)
+from core.dino_similarity.grid_extraction import DinoGridExtracorNode
```
PetrNovota: Remove the `core` package. Everything is just in this package, it doesn't make sense.

apps/dino-tracking/backend/src/core/object_selection/selection_mask_node.py
```diff
@@ -0,0 +1,95 @@
...
+        self._pending_click = (x_norm, y_norm)
+
+    def clear_selection(self) -> None:
+        self._pending_click = None
+        self._selected_mask_fs = None
+
+    def process(self, frame_msg: dai.ImgFrame, segmentation: dai.Buffer):
+        assert isinstance(segmentation, SegmentationMask)
+        frame = frame_msg.getCvFrame()
+        H_full, W_full = frame.shape[:2]
```
PetrNovota: unused variables

apps/dino-tracking/backend/src/core/object_selection/selection_mask_node.py
```diff
@@ -0,0 +1,95 @@
...
+
+        segmentation_fast_sam = segmentation.mask.astype(np.int32)
```
PetrNovota: why is this needed?

apps/dino-tracking/backend/src/core/object_selection/selection_mask_node.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if self._pending_click:
+            segment_id = self._map_click_to_segment(
+                *self._pending_click, segmentation_fast_sam
+            )
+            if segment_id is not None:
+                self._selected_mask_fs = segmentation_fast_sam == segment_id
```
PetrNovota: inconsistent naming. sometimes you use `fs` sometimes you use `fast_sam` decide and use one.
I think fast same is a NN name and we don't care about it here. we care about this being a segmentation

apps/dino-tracking/backend/src/dino_similarity/reference_vectors/reference_vector_from_selection_node.py
```diff
@@ -0,0 +1,84 @@
...
+        self,
+        mask_in: dai.Node.Output,
+        dino_in: dai.Node.Output,
+        dino_input_size: tuple[int, int],
+    ):
+        self._dino_input_size = dino_input_size
+        self.link_args(mask_in, dino_in)
+        return self
+
+    def process(self, mask_msg: dai.Buffer, dino_grid: dai.Buffer):
```
PetrNovota, line 46: `mask` is `dai.ImgFrame` not `dai.Buffer`

apps/dino-tracking/backend/src/dino_similarity/reference_vectors/reference_vector_from_selection_node.py
```diff
@@ -0,0 +1,84 @@
...
+        mask = mask_msg.getCvFrame() > 0
+
+        if self._mask_changed(mask):
```
PetrNovota, line 51: This would not be needed if you would send out a mask only when it changes. But I guess we can keep this as is

apps/dino-tracking/backend/src/dino_similarity/reference_vectors/reference_vector_from_selection_node.py
```diff
@@ -0,0 +1,84 @@
...
+        mask_in: dai.Node.Output,
+        dino_in: dai.Node.Output,
+        dino_input_size: tuple[int, int],
+    ):
+        self._dino_input_size = dino_input_size
+        self.link_args(mask_in, dino_in)
+        return self
+
+    def process(self, mask_msg: dai.Buffer, dino_grid: dai.Buffer):
+        mask = mask_msg.getCvFrame() > 0
```
PetrNovota, line 49: assert the dino_grid type

apps/dino-tracking/backend/src/main.py
FSAFTik: done

apps/dino-tracking/backend/src/main.py
FSAFTik: done

FSAFTik: done

apps/dino-tracking/backend/src/main.py
FSAFTik: done

apps/dino-tracking/backend/src/main.py
FSAFTik: done

apps/dino-tracking/backend/src/main.py
FSAFTik: done

apps/dino-tracking/backend/src/main.py
FSAFTik: done

apps/dino-tracking/backend/src/main.py
FSAFTik: Done

apps/dino-tracking/backend/src/main.py
FSAFTik: Done

apps/dino-tracking/backend/src/main.py
FSAFTik: done

apps/dino-tracking/backend/src/core/object_selection/selection_mask_node.py
FSAFTik: Removed

apps/dino-tracking/backend/src/core/object_selection/selection_mask_node.py
FSAFTik: removed

apps/dino-tracking/backend/src/dino_similarity/reference_vectors/reference_vector_from_selection_node.py
FSAFTik: Yes, but depth-ai specially requires it to be buffer (as I understand it is because it is send from other node via .out what is a Buffer)

` [2025-12-18T14:49:29Z]   File "/app/backend/src/main.py", line 103, in main
    [2025-12-18T14:49:29Z]     reference_vector = pipeline.create(ReferenceVectorFromSelection).build(
    [2025-12-18T14:49:29Z]                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    [2025-12-18T14:49:29Z]   File "/app/backend/src/dino_similarity/reference_vectors/reference_vector_from_selection_node.py", line 43, in build
    [2025-12-18T14:49:29Z]     self.link_args(mask_in, dino_in)
    [2025-12-18T14:49:29Z]   File "<string>", line 42, in link_args
    [2025-12-18T14:49:29Z] TypeError: Input 'mask_msg' cannot be linked due to incompatible message types. Input type: DatatypeEnum.ImgFrame Output type: DatatypeEnum.Buffer
`

apps/dino-tracking/backend/src/dino_similarity/reference_vectors/reference_vector_from_selection_node.py
FSAFTik: done

apps/dino-tracking/backend/src/core/object_selection/selection_mask_node.py
FSAFTik: Changed

apps/dino-tracking/backend/src/core/detections_tracking/heatmap_detection_node.py
PetrNovota: `HeatmapToDetectionsNode` also works

apps/dino-tracking/backend/src/dino_similarity/reference_vectors/reference_vector_from_selection_node.py
PetrNovota: ok, use assert then

PR #801, opened 2026-01-05
Added workflow for publishing oakapp

klemen1999:
## Purpose

Added a reusable workflow through which you can publish new apps to Hub. It can be either official app or an app specific to your team (in this case specify the full identifier)

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/.github/workflows/publish_oakapp.yaml
+++ b/.github/workflows/publish_oakapp.yaml
@@ -7,13 +7,13 @@ on:
         description: "Path to the example root containing oakapp.toml."
         type: string
         required: true
-      luxonis_official:
+      luxonis_official_identifier:
         description: "Replace com.example with com.luxonis in the identifier."
         type: boolean
-        default: false
+        default: true
         required: false
       new_identifier:
-        description: "Override identifier value in oakapp.toml."
+        description: "Override whole identifier value in oakapp.toml."
         type: string
         required: false
 
--- a/.github/ci/publish_oakapp/publish_oakapp.sh
+++ b/.github/ci/publish_oakapp/publish_oakapp.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 ROOT_DIR="${ROOT_DIR:-}"
-LUXONIS_OFFICIAL="${LUXONIS_OFFICIAL:-false}"
+LUXONIS_OFFICIAL_IDENTIFIER="${LUXONIS_OFFICIAL_IDENTIFIER:-false}"
 NEW_IDENTIFIER="${NEW_IDENTIFIER:-}"
 OAKCTL_HUB_TOKEN="${OAKCTL_HUB_TOKEN:-}"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/publish_oakapp.yaml
```diff
@@ -0,0 +1,59 @@
...
+        default: false
+        required: false
+      new_identifier:
+        description: "Override identifier value in oakapp.toml."
+        type: string
+        required: false
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
```
danilo-pejovic: This workflow will have to run on [self hosted runners ](https://github.com/luxonis/oak-examples/blob/cffe8690e3e3c7e90b14859f4d0aedfa6e4b5e03/.github/workflows/tests.yml#L125)

.github/workflows/publish_oakapp.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Install HIL
+        run: |
+          pip install hil_framework --upgrade --index-url https://__token__:${{ secrets.GITLAB_TOKEN }}@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple
+
+      - name: Publish app via HIL runner
+        env:
+          ROOT_DIR: ${{ inputs.root_dir }}
+          LUXONIS_OFFICIAL: ${{ inputs.luxonis_official }}
+          NEW_IDENTIFIER: ${{ inputs.new_identifier }}
+          OAKCTL_HUB_TOKEN: ${{ inputs.oakctl_hub_token }}
```
danilo-pejovic: I am guessing this will have to be a secret and not input? This token will live in payload - and while not printed by default before add mask step, it will be very easy for it to unintentionally leak. If one shared long lived token is possible, it would be for the best to add it to secrets - if not this can stay but lets use short lived tokens in that case.

.github/workflows/publish_oakapp.yaml
klemen1999: So I change it to `runs-on: ["self-hosted"]`? Or `runs-on: ["self-hosted", "testbed-runner"]`? Don't really know the difference to be honest.

.github/workflows/publish_oakapp.yaml
klemen1999: I see what you mean but as it stands currently I can't fit my case into either of the options because:
- Team API token is not shortlived AFAIK
- But I can't make it a secret as well because: 
  - Then anyone (or at least anyone from Luxonis team) could intentially or unintentionally push apps to official Luxonis team on Hub
  - And you also wouldn't be able to reuse this workflow to push the app to some other teams since the token would be "hardcoded" to the one of Luxonis official team

I get your concern but I'm not sure how to fix it

klemen1999: After some internal discussion I changed it so OAKCTL_HUB_TOKEN is a repo secret which means that we only allow publishing to official Luxonis team

.github/workflows/publish_oakapp.yaml
danilo-pejovic: runs-on: ["self-hosted", "testbed-runner"]

If it is just self hosted it will run on any of the runners - not specifically ones we determined to run CI jobs.

PR #806, opened 2026-01-12
Update ApplyColormap -> ApplyDepthColormap

bblazeva:
## Purpose
Update some examples which have depth visualization to use ApplyDepthColormap instead of ApplyColormap for more stable depth visualization. 

## Specification
None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


depth-measurement/calc-spatial-on-host/oakapp.toml
```diff
@@ -13,3 +13,12 @@ prepare_build_container = []
 build_steps = []
 
 entrypoint = ["bash", "-c", "python3 -u /app/main.py"]
+
+[base_image]
```
klemen1999, line 17: Out of curiosity: Why this change?

bblazeva: Because these examples install `depthai-nodes` via `git+https://` dependency now (since the `ApplyDepthColormap` is not included in a official release yet). Without the explicit` [base_image]` oakctl falls back to a default image that does not include git, which caused CI failures.
Other examples (e.g. `p2p-measurement`) already use `luxonis/oakapp-base`, which includes git, so the change aligns these examples with that working setup.

PR #826, opened 2026-02-25
Added m8-controller-box dir with examples

TamseSaso:
## Purpose

new dir for m8-controller-box with overall capabilities and examples to use with OAK4

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


m8-controller-box/README.md
klemen1999, line 1: I would suggest having a README that is more similar to other READMEs we have in other top level folders (e.g [here](https://github.com/luxonis/oak-examples/blob/feature/m8-controller-box/integrations/README.md) or [here](https://github.com/luxonis/oak-examples/blob/feature/m8-controller-box/cpp/README.md) etc.). Though since this does have some additional information about the actual interfaces we could do a merge of the two styles but I would still prefer to have this table towards the top of the README so that user can quickly jump to the example that suits them.

klemen1999, line 1: Since this PR creates a new top-level directory we need to also add it to the first page README [here](https://github.com/luxonis/oak-examples/blob/feature/m8-controller-box/README.md)

m8-controller-box/strobe-relay-example/requirements.txt
```diff
@@ -0,0 +1,7 @@
+hidapi
+git+https://github.com/luxonis/rp2040_u2if.git
+depthai
```
klemen1999, line 3: I'd prefer having at least this requirement pinned

m8-controller-box/depthai-example/requirements.txt
```diff
@@ -0,0 +1,4 @@
+depthai==3.0.0
+depthai-nodes==0.3.4
```
klemen1999, line 2: Any specific reason for having such old depthai and depthai-nodes versions?

m8-controller-box/can-example/oakapp.toml
```diff
@@ -0,0 +1,33 @@
+identifier = "com.example.m8-controller-box.can-example"
+entrypoint = ["bash", "-c", "/usr/bin/runsvdir -P /etc/service"]
+app_version = "1.0.2"
```
klemen1999, line 3: Seems strange that we start of with `1.0.2` version instead of `1.0.0` since this is an initial PR for this example

m8-controller-box/simple-example/oakapp.toml
```diff
@@ -0,0 +1,33 @@
+identifier = "com.example.m8-controller-box.simple-example"
+entrypoint = ["bash", "-c", "/usr/bin/runsvdir -P /etc/service"]
+app_version = "1.0.2"
```
klemen1999, line 3: Same comment as above

klemen1999 (commented):
Just some "stylistic" comments

PR #856, opened 2026-08-12
Migrate standalone apps to oakapp-base 1.3.0

michalvankodev:
## Summary

- Migrate the standalone examples in this change set to the `luxonis/oakapp-base` 1.3.0 image family.
- Update all standalone custom frontends to Visualizer package release 3.8.3.
- Make the custom standalone frontends reproducible in clean, constrained OAK build containers.
- Correct DINO Tracking's FastSAM parser wiring and P2P Measurement's current React ref typing.

## Base-image migration

Updates the affected Python examples to `luxonis/oakapp-base:1.3.0`, the C++ example to `1.3.0-cpp`, and the Python 3.11 frontend examples to `1.3.0-py311`. Associated frontend lockfiles are regenerated for the updated package graph.

## Visualizer frontend packages

Update the direct `@luxonis/depthai-viewer-common` and `@luxonis/depthai-pipeline-lib` dependencies from 3.7.6 to **3.8.3** for every standalone custom frontend:

- Data Collection
- DINO Tracking
- Focused Vision
- Object Volume Measurement 3D
- P2P Measurement
- People Demographics and Sentiment Analysis
- QR Tiling
- Open-Vocabulary Object Detection
- Raw Stream
- Roboflow Workflow

## Frontend build reliability

For Data Collection, DINO Tracking, Focused Vision, Object Volume Measurement 3D, P2P Measurement, People Demographics and Sentiment Analysis, and QR Tiling:

- add direct `@floating-ui/dom@1.8.0`, `@sentry/browser@8.55.2`, and `date-fns@2.30.0` dependencies required by transitive Luxonis frontend packages;
- use npm overrides to keep `react` and `react-dom` aligned to the direct React 19 versions, preventing a nested ReactDOM 18/React 19 mismatch;
- set Rollup `maxParallelFileOps: 8` to avoid `EMFILE: too many open files` in the OAK build container.

P2P Measurement additionally accepts a nullable `RefObject` in `ClickCatcher`, fixing its TypeScript build error.

## DINO Tracking correction

FastSAM parser output now feeds the mask-selection and outline paths. Its pipeline type is `dai.Buffer`, while the runtime message remains `dai.SegmentationMask`; the host node handles that contract explicitly.

## Validation

- Ran `npm install && npm run build` sequentially for all ten updated frontends with the 3.8.3 package pair; all builds pass.
- Removed `node_modules`, `dist`, and generated `styled-system` directories after every frontend build; no generated frontend artifacts remain.
- Deployed current source with `oakctl app run . --detach --invalidate` to an OAK4-S R1 (RVC4, serial `221646362`).
- Confirmed custom frontend behavior for Data Collection, DINO Tracking, Focused Vision, People Demographics and Sentiment Analysis, QR Tiling, and Open-Vocabulary Object Detection.
- Confirmed P2P Measurement and Object Volume Measurement 3D build successfully; their runtime paths correctly reject this RGB-only device because `CAM_B`/`CAM_C` are absent. A stereo-capable device is required for functional verification.

`dev-app-test.md` is intentionally excluded from this PR.

## Data Collection snap crash fix

Data Collection's snapping host node used the obsolete `SnapData(file_name=..., frame=..., detections=...)` constructor. With `depthai-nodes==0.5.0`, triggering a snap raised `TypeError: SnapData.__init__() got an unexpected keyword argument 'file_name'` inside a DepthAI host-node callback, which aborted the backend and dropped the stream.

The producer now creates a `dai.FileGroup`, adds the image/detections pair with `addImageDetectionsPair(...)`, and passes it as `SnapData(file_group=...)`. The changed Python module passes `python3 -m py_compile`.


apps/conference-demos/rgb-depth-connections/oakapp.toml
```diff
@@ -16,11 +16,16 @@ build_steps = [
 
 depthai_models = { yaml_path = "./depthai_models" }
 
+[env]
+WEBRTC__DC_MESSAGE__ORDERED = "false"
+WEBRTC__DC_MESSAGE__MAX_RETRANSMITS = "0"
+WEBRTC__DC_MESSAGE__QUEUE_POLICY = "lossy_drop_new"
+
```
klemen1999: So this is now always required for every example using `oakapp-base 1.3` image? Should we maybe have default values that are already set to this in that case so one doesn't have to always set it in the oakapp tomls manually?

michalvankodev: I was thinking about this as well. I will recommend these settings as "default" but for the period of time to consider the new protocol "stable" we've decided to make lossy type "opt in" and preserve the previous non-lossy as one that is established.

zrezke: I also think it'd be better to have these as the default. Is there any specific concern on why the new protocol wouldn't be stable?

michalvankodev: I'm personally not able to test it out on all of the types of streams we have. I have only the OAK4S. It is also very dependent on the required bandwidth for specific stream / quality. My plan forward with webrtc would be focusing on adaptable buffer size based on the stream quality / requirements. This value should be configurable by the type of the stream because it can have significant and detrimental effect when it is set to a very high value as well as very low. Goal was to have like 2-4 second buffer and then throw away and rather wait for just new stream. With a value this high. For a low quality stream with low bandwidth requirements it can be more than 40 seconds - completely defeating the purpose of having such buffer - getting 40 second delayed stream.

klemen1999 (approved):
LGTM from my side.

klemen1999 (commented):
I think we should also bump out the `app_version` of the changed apps since it is a bigger change.

PetrNovota (approved):
lgtm

PR #860, opened 2026-08-28
Feat/roboflow workflow dsp

klemen1999:
## Summary

- Migrate the Roboflow Workflow standalone app to the released OAK4 QNN runtime stack:
  - Use `luxonis/oakapp-base:1.4.0-onnxruntime`.
  - Pin `depthai==3.9.0`, `depthai-nodes==0.6.1`, and `inference==1.3.7`.
  - Run the app on the base image’s Python 3.12 runtime.
  - Pass through the Hexagon DSP/HTP devices and NPU runtime mount.

- Route Roboflow-created ONNX Runtime sessions through `depthai_nodes.runtime.onnx_qnn_session`:
  - Fix dynamic batch dimensions to static batch size 1 before QNN compilation.
  - Run QNN on HTP with fp16 and reuse cached compiled graphs.
  - Fall back to CPU by default; expose `EP=cpu` and `STRICT=1` for debugging.

- Replace the custom OpenCV generator-capture/probe path with Roboflow’s `VideoFrameProducer` integration backed by a DepthAI output queue.

- Discover workflow outputs and parameter definitions from the Roboflow API rather than performing dummy inference.

- Support runtime workflow controls:
  - Switch API key, workspace, workflow ID, and workflow parameters.
  - Rebuild DepthAI visualizer topics when output schemas change.
  - Refresh the active workflow definition from Roboflow without redeploying.
  - Surface inference pipeline health/errors in the frontend.

- Prevent workflow switching from aborting the app:
  - Quiesce `AnnotationNode` before DepthAI pipeline teardown.
  - Handle closed output queues inside the host-node callback.

- Update the custom frontend:
  - Render controls from the discovered workflow interface.
  - Add error/retry and refresh UI states.
  - Align Luxonis frontend dependencies and enforce one React/ReactDOM version.
  - Limit Rollup parallel file operations and raise the frontend build file-descriptor limit to avoid `EMFILE`.

- Simplify and update documentation:
  - Update README/agent guidance for released QNN support, runtime workflow controls, and visualization rules.

## Specification

None / not applicable

## Dependencies & Potential Impact

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

Submitted code was reviewed by a human: YES/NO

The author is taking the responsibility for the contribution: YES/NO

PR #863, opened 2026-09-07
Feat/update fe dependecies

TimurMuradian:
## Purpose
Migrates custom frontend examples from the old `@luxonis/common-fe-*` / PandaCSS stack to the new `@luxonis/ui-components` / `@luxonis/ui-utils` Tailwind-based stack.

Also keeps the examples on React 19, removes obsolete overrides/codegen setup, and updates frontend styling/imports to match the new UI libraries.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

Submitted code was reviewed by a human: YES/NO

The author is taking the responsibility for the contribution: YES/NO


caciuleanuciprian (approved):

klemen1999 (changes requested):
Examples that currently have a "normal" oakapp base image shouldn't be transfered to using `-onnxruntime` base image. The onnxruntime base image is meant to be used only in specific examples where it was already added in https://github.com/luxonis/oak-examples/pull/860 and https://github.com/luxonis/oak-examples/pull/858 PRs.
And with that in mind the oakapp.tomls also don't need specific `optional_devices`, `optional_mounts` and `allowed devices` which are specific to oakapp tomls with -onnxruntime base image. 

There is also https://github.com/luxonis/oak-examples/pull/857 PR which touches entrypoints. So it might make sense to unify so both PRs do it the same (or not change it in this one and only do it in the other one) otherwise this can get messy.

IMO I would revert most of the changes to the oakapp.tomls and mainly update their version where needed, withouth changing the type of base image that is used.

